PR #10, opened 2016-02-24
Add additional languages

gsvr:


gsvr:
@imsickofmaps Please review.

registrations/tasks.py
```diff
@@ -16,9 +16,13 @@
     "english": "eng_NG",
     "hausa": "hau_NG",
     "igbo": "ibo_NG",
+    "yoruba": "yor_NG",
+    "pidgin": "pcm_NG",
     "en": "eng_NG",
     "ha": "hau_NG",
-    "ig": "ibo_NG"
+    "ig": "ibo_NG",
```
imsickofmaps: Let's just remove the two letter ones as we don't validate them to True anyway and they're incomplete

imsickofmaps:
Minor then :+1:

PR #19, opened 2016-03-10
Using the wrong language code in the static file path.

smn:
The JSBox looks for `eng_NG` but the Django app has it under `en_NG` for registrations. 
Statechange has it under `eng_NG` so I'm going to be renaming `registrations/en_NG` to `registrations/eng_NG`.


smn:
Landing because ScaleConf, pinging @imsickofmaps so you're aware of what I'm changing as I go around fixing bunches of things.

PR #13, opened 2016-03-21
Set subscription info on registration

gsvr:
- [x] messageset_id
- [x] next_sequence_number
- [x] schedule


gsvr:
@imsickofmaps Please review.

hellomama_registration/settings.py
```diff
@@ -178,4 +178,14 @@
 CELERY_RESULT_SERIALIZER = 'json'
 CELERY_ACCEPT_CONTENT = ['json']
 
+PREBIRTH_MIN_WEEKS = 10
+PREBIRTH_MAX_WEEKS = 42
+POSTBIRTH_MIN_WEEKS = 0
+POSTBIRTH_MAX_WEEKS = 52
+
+MESSAGESET_URL = 'http://localhost:8005/api/v1/messageset/'
```
imsickofmaps: This needs an `os` set option too.

hellomama_registration/settings.py
```diff
@@ -178,4 +178,14 @@
...
+MESSAGESET_TOKEN = 'REPLACEME'
+SCHEDULE_URL = 'http://localhost:8005/api/v1/schedule/'
```
imsickofmaps: I'd prefer these were one base `stage-based-store` service URL

registrations/tasks.py
```diff
@@ -80,6 +80,80 @@ def calc_baby_age(today, baby_dob):
...
+
+def get_messageset_schedule_sequence(short_name, weeks, voice_days,
+                                     voice_times):
+    # get messageset_id
+    url = settings.MESSAGESET_URL
+    params = {'short_name': short_name}
+    headers = {'Authorization': ['Token %s' % settings.MESSAGESET_TOKEN],
+               'Content-Type': ['application/json']}
+    r = requests.get(url, params=params, headers=headers)
+    messageset_id = r.json()["id"]
```
imsickofmaps: these services calls need to be abstracted into a function that is using a client that handles retries and failures etc. For now, open an issue for the client and just move to function to unblock this issue.

registrations/tasks.py
```diff
@@ -80,6 +80,80 @@ def calc_baby_age(today, baby_dob):
...
+    else:
+        schedule_id = r.json()["default_schedule"]
+        # get schedule
+        url = settings.SCHEDULE_URL + str(schedule_id) + "/"
+        headers = {'Authorization': ['Token %s' % settings.SCHEDULE_TOKEN],
+                   'Content-Type': ['application/json']}
+        r = requests.get(url, headers=headers)
+
+    # calculate next_sequence_number
+    days_of_week = r.json()["day_of_week"]
```
imsickofmaps: needs more comments, logic is correct, but without sample data it's not clear

hellomama_registration/settings.py
```diff
@@ -178,4 +178,14 @@
 CELERY_RESULT_SERIALIZER = 'json'
 CELERY_ACCEPT_CONTENT = ['json']
 
+PREBIRTH_MIN_WEEKS = 10
```
imsickofmaps: these also need `os` options. But wrap them in `int()` because envvars are strings.

registrations/tasks.py
```diff
@@ -220,25 +296,60 @@ def create_subscriptionrequests(self, registration):
...
+        mother_msgset_id, mother_msgset_schedule, next_sequence_number =\
+            get_messageset_schedule_sequence(
+                mother_short_name, weeks, voice_days, voice_times)
         mother_sub = {
             "contact": registration.mother_id,
-            "messageset_id": 1,  # TODO
-            "next_sequence_number": 1,  # TODO
+            "messageset_id": mother_msgset_id,
+            "next_sequence_number": next_sequence_number,
             "lang": LANG_CODES[registration.data["language"]],
```
imsickofmaps: we need to not need this mapping because we're using proper lang codes. Opened https://github.com/praekelt/mama-ng-jsbox/issues/105 for the other side of this, go to using the codes as-is in the submission.

registrations/tests.py
```diff
@@ -626,8 +652,44 @@ def test_validate_baby_too_old(self):
...
+        )
+        # mock household messageset lookup
+        query_string = '?short_name=prebirth_household_text_10_42'
+        responses.add(
+            responses.GET,
+            'http://localhost:8005/api/v1/messageset/%s' % query_string,
+            json={"id": 3, "short_name": 'prebirth_household_text_10_42',
+                  "default_schedule": 3},
+            status=200, content_type='application/json',
+            match_querystring=True  # pos responses documentation!
```
imsickofmaps: I get the frustration, but this comment offers no value to the reader :smile:

imsickofmaps:
Done with first pass. Add the function stuff and I'll recheck.

gsvr:
@imsickofmaps Please have another look

registrations/tasks.py
```diff
@@ -95,60 +87,49 @@ def get_messageset_short_name(stage, recipient, msg_type, weeks):
...
-    }
-    t2 = t2_map[times]
-    t3_map = {
-        'mon_wed': "1,3",
-        'tue_thu': "2,4"
-    }
-    t3 = t3_map[days]
-    t4 = "*"
-    t5 = "*"
+def get_messageset(short_name):
```
imsickofmaps, line 100: The response will be an array of results so the right one needs returning

imsickofmaps:
Minor change then :+1:

PR #22, opened 2016-03-24
Fix webhooks to use better task wrapper

imsickofmaps:
See seed-identity-store for example


imsickofmaps:
@gsvr ready for review

gsvr:
:+1:

PR #35, opened 2016-03-30
Set welcome message on subscription metadata

imsickofmaps:
see https://github.com/praekelt/seed-stage-based-messaging/issues/9


registrations/tests.py
```diff
@@ -1111,6 +1113,8 @@ def test_friend_only_voice_mon_wed_2_5(self):
         self.assertEqual(d_friend.next_sequence_number, 5)
         self.assertEqual(d_friend.lang, "eng_NG")
         self.assertEqual(d_friend.schedule, 3)
+        self.assertEqual(d_friend.metadata["prepend_next_delivery"],
```
imsickofmaps, line 1197: @gsvr not sure about this one

imsickofmaps:
@gsvr review please

imsickofmaps:
@gsvr I've just realised I've sent the SMS as a prepend too instead of immediately as per https://github.com/praekelt/hellomama-registration/issues/36 - will resolve tomorrow, please review the rest though.

registrations/tests.py
gsvr, line 1197: I'm actually not sure either. I vaguely recall seeing a note somewhere that for this kind of registration where the same user gets the mother and household messages, that we should send only one welcome message (probably the mother's).  But I can't find it in a scope now, so I'm unsure.

gsvr:
I'm happy with it in principle.  If you want to you can land this and then we can change the welcome SMS sending in #36, otherwise we could do the work for #36 here.

imsickofmaps:
@gsvr ready for re-review

gsvr:
:+1: when travis happy

PR #42, opened 2016-04-07
Updating registration files

bruskiza:


bruskiza:
Asset push. will merge when travis.

PR #44, opened 2016-04-07
Adding public assets

bruskiza:


bruskiza:
Asset commit. cc @smn @imsickofmaps

PR #40, opened 2016-04-07
Add more helpful admin views

imsickofmaps:


imsickofmaps:
Self landing as just admin views

PR #62, opened 2016-05-18
Metric - registrations IVR vs USSD

gsvr:


gsvr:
@imsickofmaps Please review. Note I wasn't sure how the sources have been named, so we may need to change that. It may be better to just include the source name in the metric we fire.

registrations/models.py
```diff
@@ -94,6 +94,21 @@ def fire_created_metric(sender, instance, created, **kwargs):
 
 
 @receiver(post_save, sender=Registration)
+def fire_source_metric(sender, instance, created, **kwargs):
+    from .tasks import fire_metric
+    if created and 'ussd' in instance.source.name:
```
imsickofmaps: please can we user `source.user.username` instead? That's known to be a slug, and should be descriptive too. Then you won't have to have this logic in here.

imsickofmaps:
@gsvr see my comment - it'll affect a few things

gsvr:
@imsickofmaps Check last two commits, please. Wasn't clear if you wanted the metric name to be `source.user.<username>`, I just used `source.<username>`.

hellomama_registration/utils.py
```diff
@@ -206,4 +207,10 @@ def get_available_metrics():
     available_metrics.extend(settings.METRICS_REALTIME)
     available_metrics.extend(settings.METRICS_SCHEDULED)
 
+    users = User.objects.all()
+    for user in users:
+        print(user.username)
```
imsickofmaps: Stray print

hellomama_registration/utils.py
```diff
@@ -206,4 +207,10 @@ def get_available_metrics():
...
+        available_metrics.append(
+            "registrations.source.%s.sum" % user.username)
```
imsickofmaps: username's may have `@`s in them, this may cause problems

imsickofmaps: just do the Source one's - that should avoid the issue

bruskiza:
@smn @imsickofmaps poink WRT this.
I know there's a lot going on at the moment, but this has been outstanding for a while.

smn:
Is it ready for review and should I be doing the reviews? Happy to but just not sure at this point. @gsvr & @imsickofmaps is this still something you're planning on attending & wrapping up?

imsickofmaps:
Back on our radar. Will address next week.

gsvr:
@imsickofmaps Please review again.

imsickofmaps:
merge develop in 👍  once merge conflicts fixed

PR #66, opened 2016-06-27
Content upload from pathfinder HCW IVR

bruskiza:


bruskiza:
@smn @imsickofmaps fyi... as it is content... and a fairly hefty one at that.
This is the official audio from pathfinder.

Does it just still pass with an auto+1 because it's content?

smn:
thanks for uploading the audio. Are we sure the file names & file paths are correct?

bruskiza:
MOMNG-79 in JIRA was the thing.
These are mainly referenced by praekelt/mama-ng-jsbox repo.

As can be seen from the commit... it's renaming not adding.
But we will double check once this gets deployed.

bruskiza:
@smn we also have multiple other languages and apps to do... it's a bit of a process this.

smn:
:+1: on uploading, just noting that we're needing to make sure the paths & names are correct.

PR #67, opened 2016-06-28
Public IVR pathfinder upload

bruskiza:


bruskiza:
@smn same as registration.

smn:
:+1: assuming paths & names are correct.

PR #68, opened 2016-06-28
Adding public IVR training

bruskiza:


bruskiza:
@smn and will check filenames and paths correct.
Merging now... just FYI.

PR #59, opened 2016-06-30
Upgrade webhooks

imsickofmaps:
https://pypi.python.org/pypi/django-rest-hooks/1.2.2

Now properly supports `+` based, all record creators, hook sends


imsickofmaps:
Landing. Bumped package version only.

PR #70, opened 2016-07-06
Align SubscriptionRequest webhook with data expected by seed-stage-based-messaging

gsvr:
`contact` should change to `identity`


gsvr:
@imsickofmaps Please review. I wasn't sure whether I should also change the SubscriptionRequest field name to `identity`. If you don't want to have the model updated we can just revert the last commit.

imsickofmaps:
👍

PR #84, opened 2016-09-07
Add message_type metric

rudigiesler:
Add metric firing for each new registration namespaced by the metric_type

### DIFF
--- a/hellomama_registration/settings.py
+++ b/hellomama_registration/settings.py
@@ -14,7 +14,7 @@ import os
 import djcelery
 import dj_database_url
 import mimetypes
-
+from hellomama_registration.utils import MSG_TYPES
 
 # Support SVG on admin
 mimetypes.add_type("image/svg+xml", ".svg", True)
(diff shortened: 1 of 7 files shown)


rudigiesler:
Ready for review

smn:
:+1:

hellomama_registration/settings.py
```diff
@@ -14,7 +14,7 @@
 import djcelery
 import dj_database_url
 import mimetypes
-
+from hellomama_registration.utils import MSG_TYPES
```
jpaidoussi: Would it not be better to define the MSG_TYPES in settings and import them where needed rather than define them in utils? 

This might cause a circular import?

rudigiesler: I was worried about changing the functionality too much. This way it is defined statically in code, like it was previously, whereas putting it in settings makes it a dynamic thing that can be overridden in your setting file (which is possibly better), but I'm not sure what the reason was for making it static in the first place.

I don't see any reason to not put it in the settings file though

rudigiesler: Addressed in https://github.com/praekelt/hellomama-registration/pull/84/commits/608a009e76e0d7d2b35bd64323095745a5f58270

rudigiesler:
@jpaidoussi Addressed your comments. Ready for another review.

jpaidoussi:
👍

rudigiesler:
@gsvr Here's a PR to show how I've been implementing the metrics.

gsvr:
:+1:

PR #85, opened 2016-09-07
Add registration by receiver type metric

rudigiesler:
This is to add a metric firing whenever a new registration is created, namespaced by the receiver type of the registration.


rudigiesler:
@jpaidoussi Ready for review.

jpaidoussi:
👍

PR #86, opened 2016-09-08
Add message type last metric

rudigiesler:
We have the msg_type.{{type}}.sum metric, which gives us the sum of registrations per message type.

This PR adds the msg_type.{{type}}.last metric, which gives us the total amount of registrations per message type.

### DIFF
--- a/registrations/models.py
+++ b/registrations/models.py
@@ -2,6 +2,7 @@ import uuid
 
 from django.contrib.postgres.fields import JSONField
 from django.contrib.auth.models import User
+from django.core.cache import cache
 from django.db import models
 from django.db.models.signals import post_save
 from django.dispatch import receiver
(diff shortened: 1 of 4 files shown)


rudigiesler:
@jpaidoussi Ready for review.

registrations/models.py
```diff
@@ -119,16 +120,35 @@ def fire_unique_operator_metric(sender, instance, created, **kwargs):
...
-                instance.data['msg_type']),
+            "metric_name": 'registrations.msg_type.%s.sum' % msg_type,
             "metric_value": 1.0,
         })
 
+        total_key = 'registrations.msg_type.%s.last' % msg_type
+        total = cache.get(total_key)
+        if total is None:
+            total = Registration.objects.filter(
+                data__msg_type=msg_type).count()
```
jpaidoussi, line 142: Wouldn't be easier to do this from a periodic task rather than a post save signal? Is there a near-realtime requirement for this metric?

rudigiesler: I'm worried about the performance implications of doing a `count` across all registrations every 10 minutes, postgres isn't very good at counting, and doing it this way will only have us doing `count` for the first registration, and then we never have to do the `count` unless we lose the cache. The registration table is the one table that will probably have 1mil+ rows.

The metrics API will automatically aggregate things, so we don't need to worry about filling up storage space in the metrics DB.

jpaidoussi:
👍

PR #87, opened 2016-09-09
Add cache backend setting to not timeout

rudigiesler:
So it turns out that by default, django sets its cache to expire every 5 minutes. This would not be very good for our metrics caching, so we should set it to not expire.


rudigiesler:
@jpaidoussi Ready for review.

jpaidoussi:
👍

PR #90, opened 2016-09-14
Add last metric for total registrations

KaitCrawford:


KaitCrawford:
@rudigiesler ready for review

registrations/tests.py
```diff
@@ -1972,11 +1973,15 @@ def test_created_metric(self):
...
-        [request] = adapter.requests
+        [request1, request2] = adapter.requests
         self._check_request(
-            request, 'POST',
+            request1, 'POST',
             data={"registrations.created.sum": 1.0}
         )
+        self._check_request(
+            request2, 'POST',
+            data={"registrations.created.last": 1}
```
rudigiesler, line 1984: We should probably also check that if we have multiple registrations, that this has a value of 2 for the second time it gets fired, to check that it's actually incrementing properly.

rudigiesler (commented):
Looks good, just one comment about an extra test.

KaitCrawford:
@rudigiesler ready for re-review

rudigiesler (approved):
:+1:

PR #91, opened 2016-09-15
Add registration by state metrics

KaitCrawford:


KaitCrawford:
@rudigiesler ready for review

hellomama_registration/utils.py
```diff
@@ -231,3 +243,10 @@ def get_available_metrics():
                 "registrations.source.%s.sum" % source.user.username)
 
     return available_metrics
+
+
+def normalise_string(string):
+    """ Strips trailing whitespace from string, lowercases it and replaces
+        spaces with underscores
+    """
+    return ((string.strip()).lower()).replace(" ", "_")
```
rudigiesler: We probably want this to be something like `re.sub(r'\W+', '_', string)` to be absolutely sure (in case there are any strange characters)

registrations/models.py
```diff
@@ -207,6 +207,43 @@ def fire_language_metric(sender, instance, created, **kwargs):
...
+    """
+    Fires metrics for each state for each subscription, a sum metric for
+    the registrations over time, and a last metric for the total count.
+    """
+    from .tasks import fire_metric, is_valid_state
+    from hellomama_registration.utils import (get_identity, search_identities,
+                                              normalise_string)
+    if (created and instance.data and instance.data['operator_id']):
+        identity = get_identity(instance.data['operator_id'])
+        if (identity['details'] and identity['details']['state'] and
```
rudigiesler: This should probably be more something like `identity.get('details')` and `identity['details'].get('state')` to avoid value errors. (I'm assuming the identity is a `dict`)

registrations/models.py
```diff
@@ -207,6 +207,43 @@ def fire_language_metric(sender, instance, created, **kwargs):
...
+                    identity['details']['state']))):
+            state = identity['details']['state']
+            normalised_state = normalise_string(state)
+            fire_metric.apply_async(kwargs={
+                'metric_name': "registrations.state.%s.sum" % normalised_state,
+                'metric_value': 1.0,
+            })
+
+            total_key = "registrations.state.%s.last" % normalised_state
+            identities = search_identities(params={"details__state": state})
```
rudigiesler: We probably don't want to do an HTTP lookup if we don't have to, so I would rather that we put the identity search + the `Registration.object.filter(..)` in a function, and give the `get_or_incr_cache` function that function instead.

KaitCrawford:
@rudigiesler ready for re-review

smn:
:+1:

PR #92, opened 2016-09-16
Add registration by role metrics

KaitCrawford:


KaitCrawford:
@smn ready for review

registrations/models.py
```diff
@@ -251,7 +251,40 @@ def fire_state_metric(sender, instance, created, **kwargs):
 
             total = get_or_incr_cache(
                 total_key,
-                registrations_for_state(state).count)
+                registrations_for_identity_field(
+                    {"details__state": state}).count)
```
smn, line 255: Probably a silly question: does this return the `count` or does it return a reference to the [`count()`](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#django.db.models.query.QuerySet.count) function?

smn (commented):
It looks good to me but I haven't spent a _huge_ amount of time on this code base. The only thing I have a question about is the `count` reference vs `count()` call.

registrations/models.py
KaitCrawford: it returns a reference to the `count()` function
`get_or_incr_cache` takes the key to check and a method to call if the count isn't in the cache

smn: 👍

smn:
:+1: thanks @KaitCrawford

PR #94, opened 2016-09-18
Fix missing instance kwarg for manual task firing

smn:


smn:
@bruskiza ready for review

bruskiza (approved):
👍

PR #96, opened 2016-09-26
Add tools to verify (and fix) next_sequence_numbers for registrations.

smn:


smn:
ready for review @rudigiesler @KaitCrawford

smn:
I've also asked WCL to see if they can have a look.

registrations/models.py
```diff
@@ -73,6 +78,26 @@ class Registration(models.Model):
     def __str__(self):
         return str(self.id)
 
+    def get_voice_days_and_times(self):
+        return self.data.get('voice_days'), self.data.get('voice_times')
+
+    def get_preg_week(self):
```
gsvr: perhaps better named `get_week`

registrations/management/commands/verify_registration_schedule.py
```diff
@@ -0,0 +1,116 @@
...
+            "--today", dest="today", default=datetime.now(),
+            type=lambda today: datetime.strptime(today, '%Y%m%d'),
+            help=("Set the date for 'today' from which to calculate the "
+                  "next_sequence_number value. By default it will use "
+                  "datetime.now() (format YYYYMMDD)")
+        )
+        parser.add_argument(
+            '--sbm-url', dest='sbm_url', type=validate_and_return_url,
+            default=environ.get('STAGE_BASED_MESSAGING_URL'),
+            help=('The Stage Based Messaged Service to verify '
```
gsvr: `Messaged` typo

registrations/management/commands/verify_registration_schedule.py
```diff
@@ -0,0 +1,116 @@
...
+                    registration, message_set))
+
+        for request in sub_requests:
+            if (request.next_sequence_number != next_sequence_number):
+                self.log(
+                    '%s next_sequence_number is %s, should be %s' % (
+                        request.id.hex,
+                        request.next_sequence_number, next_sequence_number))
+                if apply_fix:
+                    request.next_sequence_number = next_sequence_number
```
gsvr: It won't affect anything currently with the few messagesets that can be subscribed to, but I would set the `messageset` and `schedule` here as well in case they have changed based on the recalculation. And though it is highly unlikely that the `messageset` or `schedule` would change but not the `next_sequence_number`, I'd check for that possibility as well in the `if` in line 99.

registrations/management/commands/verify_registration_schedule.py
```diff
@@ -0,0 +1,116 @@
...
+                'environment variable or --sbm-url is set.')
+
+        if not sbm_token:
+            raise CommandError(
+                'Please make sure either the STAGE_BASED_MESSAGING_TOKEN '
+                'environment variable or --sbm-token is set.')
+
+        client = StageBasedMessagingApiClient(sbm_token, sbm_url)
+        subscriptions = client.get_subscriptions({
+            'identity': registration.mother_id,
```
gsvr: This is catering for the subscriptions if only the mother wants to receive messages, but I can't see anything for the househould subscriptions - I think this may have been overlooked due to the lacking documentation. See here: https://github.com/praekelt/hellomama-registration/blob/develop/registrations/tasks.py#L250-L277 for household subscription creation.

registrations/models.py
```diff
@@ -73,6 +78,26 @@ class Registration(models.Model):
...
+        week = self.data.get("preg_week") or self.data.get("baby_age")
+        if week is not None:
+            return week
+        raise RegistrationException(
+            'Neither preg_week or baby_age are specified for %s.' % (self,))
+
+    def get_subscription_requests(self):
+        return SubscriptionRequest.objects.filter(identity=self.mother_id)
```
gsvr: This is also limited to the mother's subscriptions.

gsvr (changes requested):

registrations/models.py
smn: Considering this is on the `Registration` model calling it `get_week` could lead to some confusion as to it returning the week of what, the registration or the pregnancy?

I'm happy to rename it but I think `get_week()` adds to much ambiguity and isn't an improvement.

registrations/models.py
smn: @gsvr could you elaborate on this one a bit please? I'm not sure what you're implying as it already is currently limited to the mother?

registrations/models.py
gsvr: Agreed, maybe `get_weeks_pregnant_or_age`? Or something similar.

registrations/models.py
gsvr: Ties in with my comment in `commands/verify_registration_schedule` - we can discuss in a call if that doesn't clarify enough.

registrations/management/commands/verify_registration_schedule.py
smn: `household` is a command line option [here](https://github.com/praekelt/hellomama-registration/pull/96/files#diff-f77d84ddc442d43bf67d59efa98f3401R29).

Are you saying we need to filter on `household` subscriptions there too?

registrations/management/commands/verify_registration_schedule.py
smn: Updated.

smn:
@gsvr ready for review again

smn:
@gsvr ready for review again.

registrations/management/commands/verify_registration_schedule.py
```diff
@@ -0,0 +1,161 @@
...
+            'identity': receiver_id,
+        })
+
+        if subscriptions['count']:
+            raise CommandError(
+                'Subscriptions exist for %s.' % (registration,))
+
+        weeks_estimate = registration.estimate_current_preg_weeks(today=today)
+        voice_days, voice_times = registration.get_voice_days_and_times()
+        mother_short_name = utils.get_messageset_short_name(
```
gsvr: Please change `mother_short_name` to `messageset_short_name` since it's not mother specific any more.

gsvr (approved):
Just one small change requested.

The only thing I would note is that this tool will only work for prebirth registrations. It will require expansion for checking postbirth registrations if those are ever re-activated.

smn:
@gsvr how would I recognise post-birth registrations? Would like to raise an error for those specifically.

smn:
ah `registration.stage` has that information.

PR #97, opened 2016-09-27
harden subscription firing

smn:


smn:
@jpaidoussi @rudigiesler @bruskiza hardened this management command to not do anything if _any_ subscriptions already exist for a given identity. This may be overly cautious as the subscriptions available may not be the ones missing but I'd prefer erring on the side of caution at this point.

smn:
this is ready for review.

registrations/management/commands/verify_registration_schedule.py
```diff
@@ -3,16 +3,11 @@
...
 from seed_services_client import StageBasedMessagingApiClient
 
 from registrations.models import Registration
 from hellomama_registration import utils
-
-
-def validate_and_return_url(url):
-    URLValidator()(url)
-    return url
+from ._utils import validate_and_return_url
```
smn, line 10: underscored because it is in the management commands folder but [not a management command](https://docs.djangoproject.com/en/1.10/howto/custom-management-commands/#module-django.core.management)

rudigiesler:
Looks good to me, :+1:

PR #100, opened 2016-10-27
Add metric repopulation

rudigiesler:
We need to be able to repopulate metrics for older registration data.


rudigiesler:
In order to reduce the size of this PR, I've decided to add the framework for metric repopulation here, and to implement the specific metric calculations in another PR.

rudigiesler:
@KaitCrawford Ready for review.

rudigiesler:
@smn The failing tests here are due to changes that you made in https://github.com/praekelt/hellomama-registration/pull/96 , not sure why it's failing, it's failing for me on develop too.

smn:
@rudigiesler perhaps a date isn't being mocked properly somewhere and it worked for the date the PR got merged?

rudigiesler:
@smn Looking into it, it seems like we're mocking the date correctly with the `today` parameter, but the `mk_registration_at_week` is working with `datetime.now()`. Adding a `today` variable to it seems to fix it, but I'm not sure that I know enough about the system to know whether it's an issue, or whether it's just the test that's wrong. Here's the diff for the fix: https://gist.github.com/b22dd245f6287af29dbc9296e38b1398

smn:
I suspect that's it @rudigiesler, thanks! 🍰

rudigiesler:
Screenshots

Added button:
image

Added form:
image

registrations/graphite.py
```diff
@@ -0,0 +1,79 @@
...
+            now = datetime.utcnow()
+        finish = now
+
+        for r in self.retentions:
+            beginning = now
+            for start, end in r.get_buckets(now=now, finish=finish):
+                beginning = min(beginning, start)
+                yield (start, end)
+            # The next retention should end where this one started
+            finish = beginning
```
KaitCrawford, line 81: I'm struggling to wrap my head around this code block. I don't see how the time buckets returned for smaller durations aren't overlapped by the buckets returned for longer durations. Perhaps it would help if the variable names were more descriptive? Like `beggining` -> `retention_start` and `start` -> `bucket_start` or something like that?

registrations/graphite.py
```diff
@@ -0,0 +1,79 @@
...
+
+        while start < finish:
+            end = start + self.precision
+            if end > finish:
+                end = finish
+            yield (start, end)
+            start = end
+
+
+class GraphiteRetentions(object):
```
KaitCrawford: I think this would be better off named `RetentionScheme` or something that isn't so similar to the class above it.

KaitCrawford (commented):
@rudigiesler I'm quite confused by all this bucketing. I thought Graphite handled that for us and we just had to tell it when the event occurred?

registrations/graphite.py
rudigiesler: I'm definitely up for suggestions on better naming. Basically, it gets all the buckets from the first retention scheme (eg. 1s:1h).

It then looks for the earliest start date out of all the buckets for that retention scheme (in our example, that would be 1 hour ago).

It then gets the buckets for the next retention scheme (eg 1h:2h), making sure to stop before the earliest start date of the previous retention scheme's buckets. (in our example, we'll just get a bucket for (2h ago, 1h ago), since the time from one hour ago to now was already covered by the previous retention scheme).

Doing it this way means that we never have any overlapping buckets.

rudigiesler:
@KaitCrawford No graphite doesn't do this for us. The vumi metrics workers do this for us, but they do this by collecting the metrics that we give it, and when the timeframe for our bucket ends, it then fires the aggregated value to Graphite. If we send graphite more than one value for a single bucket, Graphite just overwrites the existing value with the one we're sending it.

What Graphite does do is aggregate values as they transition from one retention to another (eg. if we had  `1m:1h,1h:8h,4h:1d`, then all data older than 1 hour would automatically be aggregated together in a single value for that hour for the next 8 hours, then when that data is older than 8 hours it gets aggregated into a single value for for every 4 hours for the next day, then when the data is older than a day it gets removed.)

registrations/graphite.py
rudigiesler: Resolved with https://github.com/praekelt/hellomama-registration/pull/100/commits/a681d1c3c23638bddea707aa4b3b8a5f9f493b77

registrations/graphite.py
KaitCrawford: Ah ok. I see now where I got lost. Perhaps some step by step comments would help people follow it in future?

KaitCrawford:
Thanks for the explanation @rudigiesler
I think some comments or different variable names would make the retention bucket code easier to read but you can decide on that. I'm :+1: overall.

PR #102, opened 2016-11-01
Switch to one long task instead of lots of subtasks

rudigiesler:
When running the repopulate metrics tasks with popular production values, it takes 4.5GB of memory before it starts queueing the subtasks. Even if we don't collect all the tasks before hand, and just start queuing them as we generate them, it still takes too much memory.

This PR combines the two tasks to generate metrics into one long running task.


KaitCrawford:
:+1:

PR #103, opened 2016-11-02
Reuse amqp connection for metric sending

rudigiesler:
Because we previously had each metric as a separate celery task, we needed to create a new connection for every metric. Turns out doing this when running everything in one task causes connection issues, so the goal is to streamline the connection by creating one connection, and sending all the metrics over that amqp connection.

### DIFF
--- a/registrations/metrics.py
+++ b/registrations/metrics.py
@@ -1,5 +1,4 @@
 import pika
-
 from django.db.models import Count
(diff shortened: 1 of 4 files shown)


rudigiesler:
@KaitCrawford Ready for review

KaitCrawford:
:+1:

PR #104, opened 2016-11-11
State change: Language change metric

erikh360:
This PR adds the "change.language.sum" metric.

### DIFF
--- a/changes/models.py
+++ b/changes/models.py
@@ -7,7 +7,7 @@ from django.contrib.auth.models import User
 from django.contrib.postgres.fields import JSONField
 from django.utils.encoding import python_2_unicode_compatible
 
-from registrations.models import Source
+from registrations.models import Source, get_or_incr_cache
 
 
 @python_2_unicode_compatible
@@ -68,4 +68,13 @@ def fire_language_change_metric(sender, instance, created, **kwargs):
         fire_metric.apply_async(kwargs={
             "metric_name": 'change.language.sum',
             "metric_value": 1.0
+        })
+
+        total_key = 'change.language.total.last'
+        total = get_or_incr_cache(
+            total_key,
+            Change.objects.filter(action='change_language').count)
+        fire_metric.apply_async(kwargs={
+            'metric_name': total_key,
+            'metric_value': total,
         })
--- a/changes/models.py
+++ b/changes/models.py
@@ -66,11 +66,11 @@ def fire_language_change_metric(sender, instance, created, **kwargs):
     from registrations.tasks import fire_metric
     if created and instance.action == 'change_language':
         fire_metric.apply_async(kwargs={
-            "metric_name": 'change.language.sum',
+            "metric_name": 'registrations.change.language.sum',
             "metric_value": 1.0
         })
 
-        total_key = 'change.language.total.last'
+        total_key = 'registrations.change.language.total.last'
         total = get_or_incr_cache(
             total_key,
             Change.objects.filter(action='change_language').count)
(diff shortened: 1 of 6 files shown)


erikh360:
@rudigiesler please review.

changes/models.py
```diff
@@ -59,3 +59,13 @@ def change_post_save(sender, instance, created, **kwargs):
...
+
+
+@receiver(post_save, sender=Change)
+def fire_language_change_metric(sender, instance, created, **kwargs):
+    from registrations.tasks import fire_metric
+    if created and instance.action == 'change_language':
+        fire_metric.apply_async(kwargs={
+            "metric_name": 'change.language.sum',
+            "metric_value": 1.0
+        })
```
rudigiesler, line 71: I think we also want a `.total.last` metric here for the total amount of language changes (for most of the metrics tickets, unless specified, it's implied that we want both types of metrics).

registrations/metrics.py
```diff
@@ -180,6 +181,14 @@ def registrations_source_sum(self, user, start, end):
             .filter(source__user__username=user)\
             .count()
 
+    def change_language_sum(self, user, start, end):
+        return Change.objects\
+            .filter(created_at__gt=start)\
+            .filter(created_at__lte=end)\
+            .filter(source__user__username=user)\
+            .filter(action='change_language')\
+            .count()
```
rudigiesler, line 195: Do we also maybe want to add tests for this function? Since I'm not sure that it's 100% correct, because these functions only get sent `start` and `end`, no other arguments.

We should also create a `Change` model in this test: https://github.com/praekelt/hellomama-registration/blob/develop/registrations/test_metrics.py#L484-L493 , to ensure that the function is named correctly for the system to find it.

erikh360:
@KaitCrawford please review.
I added the total.last with tests that Rudi suggested, I also added tests for the metric generator functions.

hellomama_registration/settings.py
```diff
@@ -237,6 +237,8 @@
     'registrations.created.sum',
     'registrations.created.total.last',
     'registrations.unique_operators.sum',
+    'change.language.sum',
+    'change.language.total.last'
```
KaitCrawford: Off the top of my head, I feel like these should have the `registrations` prefix like the `created` metrics above

KaitCrawford: I think it might be useful to track the metrics based on the language they were changing from/to. I'm not sure though. Let's see if we can corner Jacqui and chat to her about it.

registrations/test_metrics.py
```diff
@@ -9,14 +9,112 @@
...
+            "Change model still has post_save listeners. Make sure"
+            " helpers removed them properly in earlier tests.")
+        post_save.connect(receiver=change_post_save,
+                          sender=Change)
+        post_save.connect(receiver=fire_language_change_metric,
+                          sender=Change)
+        post_save.connect(receiver=model_saved,
+                          dispatch_uid='instance-saved-hook')
+
+    def _replace_post_save_hooks_registration(self):
```
KaitCrawford: If I'm not mistaken, this class inherits `registrations.tests.AuthenticatedAPITestCase` that means this method isn't needed because it is exactly the same as the one here https://github.com/praekelt/hellomama-registration/blob/6a9ffcbf92b06228a51a7626c91ae4a9fbb18a13/registrations/tests.py#L184-L212

KaitCrawford (commented):
@erikh360 I've left a few comments. I'm concerned that we're only tracking the overall language changes rather than the details of the languages that users swapped between. Let's try and chat to Jacqui to find out if they need more detail.

erikh360:
Hi @KaitCrawford please review again.

hellomama_registration/settings.py
KaitCrawford: We confirmed with Jacqui that the current work is sufficient since the metrics will only be used to determine if the feature is being used.

KaitCrawford:
:+1:

PR #109, opened 2016-11-17
Add management command for repopulating subscriptions

KaitCrawford:
There have been a number of bugs that have prevented registrations from being converted into subscriptions (the most recent being incorrect message set names). These bugs have been fixed and this PR adds a management command to re-run the validation on registrations that don't have subscriptions. This will result in the subscriptions being created.

### DIFF
--- a/registrations/management/commands/repopulate_subscriptions.py
+++ b/registrations/management/commands/repopulate_subscriptions.py
@@ -53,8 +53,9 @@ class Command(BaseCommand):
         registrations = Registration.objects.filter(validated=True)
 
         for reg in registrations:
+            skip = False
             requests = reg.get_subscription_requests()
-            if requests.count() > 0:
+            if requests.exists():
                 continue
             for req in requests:
                 if check_subscription and self.count_subscriptions(client,
@@ -63,7 +64,10 @@ class Command(BaseCommand):
                               'already has subscription (identity: %s). '
                               'Skipping.')
                              % (reg.pk, reg.mother_id))
-                    continue
+                    skip = True
+                    break
+            if skip:
+                continue
 
             """
             validate_registration() ensures no invalid registrations get
--- a/registrations/management/commands/repopulate_subscriptions.py
+++ b/registrations/management/commands/repopulate_subscriptions.py
@@ -74,7 +74,7 @@ class Command(BaseCommand):
             subscriptions and creates the Subscription Request
             """
             output = validate_registration.apply_async(
-                registration_id=str(reg.id))
+                kwargs={"registration_id": str(reg.id)})
             output = output + " (%s)"
             self.log(output % (reg.mother_id))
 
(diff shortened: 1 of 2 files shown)


KaitCrawford:
@smn  please review

registrations/management/commands/repopulate_subscriptions.py
```diff
@@ -0,0 +1,83 @@
...
+                              'already has subscription (identity: %s). '
+                              'Skipping.')
+                             % (reg.pk, reg.mother_id))
+                    continue
+
+            """
+            validate_registration() ensures no invalid registrations get
+            subscriptions and creates the Subscription Request
+            """
+            output = validate_registration(registration_id=str(reg.id))
```
rudigiesler: I think this needs to be indented to be in the for loop.

rudigiesler (commented):
Could we also have a unit test for this management command? You can use https://docs.djangoproject.com/en/1.10/ref/django-admin/#django.core.management.call_command

registrations/management/commands/repopulate_subscriptions.py
```diff
@@ -0,0 +1,83 @@
...
+                    'Please make sure either the STAGE_BASED_MESSAGING_TOKEN '
+                    'environment variable or --sbm-token is set.')
+            client = StageBasedMessagingApiClient(sbm_token, sbm_url)
+
+        registrations = Registration.objects.filter(validated=True)
+
+        for reg in registrations:
+            requests = reg.get_subscription_requests()
+            if requests.count() > 0:
+                continue
```
smn, line 64: PostgreSQL is _phenomenally_ bad at counting things. It's always better using `.exists()` than `count()` as the existence for at least 1 record is always cheaper than doing a count of exactly how many there are.

registrations/management/commands/repopulate_subscriptions.py
```diff
@@ -0,0 +1,83 @@
...
+            for req in requests:
+                if check_subscription and self.count_subscriptions(client,
+                                                                   req):
+                    self.log(('Registration %s without Subscription Requests '
+                              'already has subscription (identity: %s). '
+                              'Skipping.')
+                             % (reg.pk, reg.mother_id))
+                    continue
```
smn: I probably should now this but I'm not sure if this breaks us out of both `for` loops here.

smn (commented):
Looks good, just let a few minor comments. It would be good to have a test for this. Shout if you need help getting started on that, Django has a `call_command` method that you can use to test these kinds of things.

KaitCrawford:
@rudigiesler @smn ready for re-review

registrations/test_commands.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class ManagementCommandsTests(AuthenticatedAPITestCase):
+    def test_command_requires_sbm_url(self):
+        with self.assertRaises(management.CommandError) as ce:
+            management.call_command("repopulate_subscriptions")
+        self.assertEqual(
+            str(ce.exception), "Please make sure either the "
+            "STAGE_BASED_MESSAGING_URL environment variable or --sbm-url is "
+            "set.")
```
smn, line 20: hah, never thought of doing this. Excellent :) :cake:

smn (approved):

PR #112, opened 2016-11-25
Regenerate personnel code from admin

KaitCrawford:


KaitCrawford:
@rudigiesler ready for review

uniqueids/admin.py
```diff
@@ -1,6 +1,39 @@
...
+    def regenerate_personnel_code(self, request, queryset):
+        created = 0
+        skipped = 0
+        for record in queryset:
+            if record.write_to != "personnel_code":
+                skipped += 1
+                continue
+            Record.objects.create(
+                identity=record.identity, write_to="personnel_code",
+                length=record.length, created_by=request.user)
```
rudigiesler: I'm not sure if we want to create new codes? I think it might be better for this to be idempotent, so all that it does is resend existing codes.

I'm thinking of the case of where say we've been creating health IDs for a month, but only in the past week have we been sending the codes. If we select all from the past month to resend, then it will disable the codes that were sent in the last week. I think it would cause less confusion in those cases to just resend the SMS with the code, rather than sending them a new code.

uniqueids/admin.py
```diff
@@ -1,6 +1,40 @@
...
+    list_display = [
+        "id", "identity", "write_to", "created_at", "updated_at"]
+    list_filter = ["write_to", "created_at"]
+    search_fields = ["identity", "write_to"]
+    actions = ["resend_personnel_code"]
+
+    def resend_personnel_code(self, request, queryset):
+        created = 0
+        skipped = 0
+        for record in queryset:
```
miltontony: Please use `.iterator()`

uniqueids/admin.py
```diff
@@ -1,6 +1,40 @@
...
+            if record.write_to != "personnel_code":
```
miltontony: I think you can just do `for record in queryset.exclude(write_to='personnel_code')` You won't have your count though.

KaitCrawford:
@miltontony ready for re-review

uniqueids/admin.py
```diff
@@ -1,6 +1,30 @@
...
+                "identity": str(record.identity),
+                "personnel_code": record.id})
+            created += 1
+        if created == 1:
+            created_text = "%s Record was" % created
+        else:
+            created_text = "%s Records were" % created
+        self.message_user(request, "%s resent." % created_text)
+
+    resend_personnel_code.short_description = "Send code by SMS (personnel "\
```
miltontony: This is a bit awkward. You can do the same thing by using `()`
```python
resend_personnel_code.short_description = (
    "Send code by SMS (personnel code only)")
```

miltontony:
small last request, but otherwise 👍 🍰

PR #117, opened 2017-01-06
Optout metrics

erikh360:
- Opt out hook receiver function
- Opt out by receiver type metric
- Opt out by reason metric
- Opt out by message type metric
- Opt out by source metric

### DIFF
--- a/changes/views.py
+++ b/changes/views.py
@@ -5,6 +5,7 @@ from rest_framework import viewsets, mixins, generics, filters
 from rest_framework.permissions import IsAuthenticated
 from .serializers import ChangeSerializer
 from django.http import JsonResponse
+from django.conf import settings
 from hellomama_registration import utils
 
 import six
@@ -86,6 +87,9 @@ class ReceiveIdentityStoreOptout(mixins.CreateModelMixin,
 def fire_optout_reason_metric(reason):
     from registrations.tasks import fire_metric
 
+    if reason not in settings.OPTOUT_REASONS:
+        reason = 'other'
+
     fire_metric.apply_async(kwargs={
         "metric_name": 'optout.reason.%s.sum' % reason,
         "metric_value": 1.0
@@ -111,4 +115,24 @@ def fire_optout_receiver_type_metric(msg_receiver):
     fire_metric.apply_async(kwargs={
         "metric_name": 'optout.receiver_type.%s.sum' % msg_receiver,
         "metric_value": 1.0
+    })
+
+    def search_optouts_receiver_type():
+        result = utils.search_optouts()
+
+        identities = []
+        for data in result:
+            identities.append(data['identity'])
+
+        return Registration.objects.filter(
+                    mother_id__in=identities,
+                    data__msg_receiver=msg_receiver).count()
+
+    total_key = 'optout.receiver_type.%s.total.last' % msg_receiver
+    total = get_or_incr_cache(
+        total_key,
+        search_optouts_receiver_type)
+    fire_metric.apply_async(kwargs={
+        'metric_name': total_key,
+        'metric_value': total,
     })
--- a/changes/views.py
+++ b/changes/views.py
@@ -68,7 +68,6 @@ class ReceiveIdentityStoreOptout(mixins.CreateModelMixin,
 
         try:
             identity_id = data['identity']
-            # optout_type = data['optout_type']
             optout_reason = data['optout_reason']
         except KeyError as e:
             return JsonResponse({'reason': '"identity", "optout_type" and '
@@ -80,6 +79,9 @@ class ReceiveIdentityStoreOptout(mixins.CreateModelMixin,
             fire_optout_receiver_type_metric(registration.data['msg_receiver'])
 
         fire_optout_reason_metric(optout_reason)
+
+        if registration.data.get('msg_type'):
+            fire_optout_message_type_metric(registration.data['msg_type'])
 
         return JsonResponse({})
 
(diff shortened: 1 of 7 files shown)


erikh360:
@rudigiesler please review.

changes/views.py
```diff
@@ -46,3 +52,147 @@ class ChangeGetViewSet(viewsets.ReadOnlyModelViewSet):
...
+        except ValueError as e:
+            return JsonResponse({'reason': "JSON decode error",
+                                'details': six.text_type(e)}, status=400)
+
+        try:
+            identity_id = data['identity']
+            optout_reason = data['optout_reason']
+            optout_source = data['optout_source']
+        except KeyError as e:
+            return JsonResponse({'reason': '"identity", "optout_type", '
```
rudigiesler: I don't see `optout_type` above?

changes/views.py
```diff
@@ -46,3 +52,147 @@ class ChangeGetViewSet(viewsets.ReadOnlyModelViewSet):
...
+                                 '"optout_reason" and "optout_source" must be '
+                                 'specified.'
+                                 }, status=400)
+
+        registration = Registration.objects.get(mother_id=identity_id)
```
rudigiesler: I think it's possible for an identity to have more than 1 registration, or no registrations, we should probably have handling for those cases.

changes/views.py
```diff
@@ -46,3 +52,147 @@ class ChangeGetViewSet(viewsets.ReadOnlyModelViewSet):
...
+        reason = 'other'
+
+    fire_metric.apply_async(kwargs={
+        "metric_name": 'optout.reason.%s.sum' % reason,
+        "metric_value": 1.0
+    })
+
+    def search_optouts_reason():
+        result = utils.search_optouts({"reason": reason})
+        return len(list(result))
```
rudigiesler: This will cause us to load the entire list of results in memory before counting it. Maybe we can do something like `sum(1 for r in result)` to count it without loading it all in memory?

changes/views.py
```diff
@@ -46,3 +52,147 @@ class ChangeGetViewSet(viewsets.ReadOnlyModelViewSet):
...
+    source_short = source.split('_')[0]
+
+    fire_metric.apply_async(kwargs={
+        "metric_name": 'optout.source.%s.sum' % source_short,
+        "metric_value": 1.0
+    })
+
+    def search_optouts_source():
+        result = utils.search_optouts({"request_source": source})
+        return len(list(result))
```
rudigiesler: Same here about loading the entire result set into memory.

changes/views.py
```diff
@@ -46,3 +52,147 @@ class ChangeGetViewSet(viewsets.ReadOnlyModelViewSet):
...
+        "metric_name": 'optout.receiver_type.%s.sum' % msg_receiver,
+        "metric_value": 1.0
+    })
+
+    def search_optouts_receiver_type():
+        result = utils.search_optouts()
+
+        identities = []
+        for data in result:
+            identities.append(data['identity'])
```
rudigiesler: We can probably use list comprehension here, and use a set instead of a list to ensure that we don't have duplicates. Like `identities = set(data['identity'] for data in result)`

changes/views.py
```diff
@@ -46,3 +52,147 @@ class ChangeGetViewSet(viewsets.ReadOnlyModelViewSet):
...
+        "metric_name": 'optout.msg_type.%s.sum' % msg_type,
+        "metric_value": 1.0
+    })
+
+    def search_optouts_message_type():
+        result = utils.search_optouts()
+
+        identities = []
+        for data in result:
+            identities.append(data['identity'])
```
rudigiesler: Same comment here as the receiver type metric

hellomama_registration/utils.py
```diff
@@ -92,6 +93,24 @@ def patch_identity(identity, data):
     return r.json()
 
 
+def search_optouts(params=None):
```
rudigiesler, line 96: Not strictly necessary, but if we have `**params` here, then we can call this function `search_optouts(reason='other', request_source='ussd')` instead of `search_optouts(params={'reason': "other", 'request_source': "ussd"})`

erikh360:
@rudigiesler ready for re-review, thanks.

registrations/metrics.py
```diff
@@ -233,6 +271,90 @@ def registrations_change_messaging_total_last(self, start, end):
...
 
+    def optout_msg_type_sum(self, msg_type, start, end):
+        result = utils.search_optouts({
+            "created_at__gt": start,
+            "created_at__lte": end,
+        })
+
+        identities = []
+        for data in result:
+            identities.append(data['identity'])
```
rudigiesler: We can probably do the same `set(data['identity'] for data in result)` that we did for the metrics firing here in the metrics regeneration (for all the metric types here).

registrations/metrics.py
```diff
@@ -233,6 +271,90 @@ def registrations_change_messaging_total_last(self, start, end):
...
+                    mother_id__in=identities,
+                    data__msg_receiver=receiver_type).count()
+
+    def optout_reason_sum(self, reason, start, end):
+        result = utils.search_optouts({
+            "reason": reason,
+            "created_at__gt": start,
+            "created_at__lte": end,
+        })
+        return len(list(result))
```
rudigiesler: Same here for using `sum(1 for r in result)` instead of `len(list(result))`, and the ones below this

rudigiesler (commented):
Just two small comments, looking good otherwise :)

rudigiesler (approved):

PR #118, opened 2017-01-12
Add http retries for metric repopulation

rudigiesler:
Currently, the metric repopulation task fails if one of the HTTP requests to one of the external services fails. We should retry failed requests a number of times, and if we still can't get through, we should just skip that metric instead of failing the whole task.


rudigiesler:
Ready for review.

erikh360 (approved):

PR #119, opened 2017-01-13
Combine management commands for fixing subscriptions into one.

KaitCrawford:


KaitCrawford:
@miltontony, @jpaidoussi, @erikh360 or @rudigiesler please review

registrations/management/commands/fix_registration_subscriptions.py
```diff
@@ -0,0 +1,351 @@
...
+                 (msgset_subscription['id'],
+                  msgset_subscription["next_sequence_number"],
+                  sub_request.next_sequence_number))
+        if apply_fix:
+            client.update_subscription(
+                msgset_subscription['id'],
+                {'next_sequence_number': sub_request.next_sequence_number})
+            self.log('Updated subscription %s. Set next_sequence_number to %s'
+                     % (msgset_subscription['id'],
+                        sub_request.next_sequence_number))
```
rudigiesler, line 299: I'm a little confused by this bit.

If the subscription has just been created, then we should expect the `next_sequence_number` on the subscription request to match the actual subscription.

But, if the subscription has run even once, then we should expect the two `next_sequence_number`s to not match. But it seems here for that case, we will be resetting the user to the beginning of the messageset again.

Should we not be comparing the subscriptionrequest's `next_sequence_number + calculated_number_of_runs` to the subscription's `next_sequence_number`?

rudigiesler (commented):
Just the one question from me.

registrations/management/commands/fix_registration_subscriptions.py
KaitCrawford: @rudigiesler the code to check the subscription request will run before this and change the `next_sequence_number` of the request if it doesn't match what we expect. so the value on the subscription request should be what we expect.

rudigiesler: Ah, I understand now.

rudigiesler (approved):

PR #120, opened 2017-01-16
Management command to populate initial sequence in SBM

erikh360:
We are adding a new field in the seed-stage-based-messaging for the initial_sequence_number, this command will populate it.

### DIFF
--- a/registrations/management/commands/update_initial_sequence.py
+++ b/registrations/management/commands/update_initial_sequence.py
@@ -1,0 +1,83 @@
+from os import environ
+
+from django.core.management.base import BaseCommand, CommandError
+
+from registrations.models import SubscriptionRequest
+
+from seed_services_client import StageBasedMessagingApiClient
+
+from ._utils import validate_and_return_url
+
+
+class Command(BaseCommand):
+    help = ("This command will loop all subscription requests and find the "
+            "corresponding subscription in SBM and update the "
+            "initial_sequence_number field, we need this to fast forward the "
+            "subscription.")
+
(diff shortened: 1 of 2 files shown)


registrations/management/commands/update_initial_sequence.py
```diff
@@ -0,0 +1,83 @@
...
+            raise CommandError(
+                'Please make sure either the STAGE_BASED_MESSAGING_TOKEN '
+                'environment variable or --sbm-token is set.')
+
+        sbm_client = StageBasedMessagingApiClient(sbm_token, sbm_url)
+
+        sub_requests = SubscriptionRequest.objects.all()
+
+        updated = 0
+        for sub_request in sub_requests:
```
KaitCrawford, line 48: This is going to be a huge QuerySet. Use .iterator() instead https://docs.djangoproject.com/en/1.10/ref/models/querysets/#iterator

registrations/management/commands/update_initial_sequence.py
```diff
@@ -0,0 +1,83 @@
...
+            subscriptions = sbm_client.get_subscriptions({
+                'identity': sub_request.identity,
+                'created_at__gt': sub_request.created_at
```
KaitCrawford: I think we should also make sure the `message_set` is the same for the request and the subscription

KaitCrawford (commented):
@erikh360 Two small comments.
This is going to be a huge thing to run since it loops through ALL subscription requests. we're going to have to be careful about what time we run it at

PR #123, opened 2017-01-17
Fix for when the mother is not one of the receivers

erikh360:


changes/tests.py
```diff
@@ -198,7 +198,8 @@ def make_registration_friend_only(self):
                 "msg_receiver": "friend_only",
                 # data added during validation
                 "reg_type": "hw_pre",
-                "preg_week": "11"
+                "preg_week": "11",
+                "msg_receiver": "friend_only",
```
KaitCrawford: @erikh360 isn't this line the same as 198 above?

KaitCrawford (approved):
@erikh360 super small duplicate line. otherwise :+1:

changes/tests.py
erikh360: Oops, good eye! fixing now.

PR #125, opened 2017-02-09
Fix date for management command test

KaitCrawford:
This test was failing because the output differs based on the current date. The command being tested has provision for setting the date to use, it just wasn't being called that way in the test.


miltontony:
👍

PR #126, opened 2017-02-13
Port management command for report generation

KaitCrawford:
This is the management command from the CI that generates the HelloMama reports.
(https://github.com/praekelt/seed-control-interface/blob/develop/ci/management/commands/generate_reports.py)
It has been adjusted to retrieve Change and Registration objects from the database rather than from the api endpoints.


rudigiesler (approved):
:+1:, looks good to me. Just to confirm, is this the latest that we have in the CI?

KaitCrawford:
@rudigiesler yeah, I made sure to include the changes to how the message sender is queried. That was the only thing that changed since I started working on it.

PR #130, opened 2017-03-06
Use json.dumps(), boolean was not working

erikh360:


rudigiesler (approved):
Since we're using python `requests`, we can just have `json=data` in the `session.patch`, ie `session.patch(url, json=data, headers=headers)`. Either way, looks good to me.

rudigiesler:
One last thing, how come no tests picked this up? Do we maybe need a test for this?

erikh360:
The patch requests are there, it's just not validating the json payload:

`
# mock patch subscription request
        responses.add(
            responses.PATCH,
            'http://localhost:8005/api/v1/subscriptions/%s/' % subscription_id,
            json={"active": False},
            status=200, content_type='application/json',
        )
`

rudigiesler:
Ah okay. I see. Is the payload maybe something we would want to validate? If not, then :+1: on this PR.

erikh360:
I think we should actually be changing this to use the `seed-services-client`, for now I'm merging this.

PR #131, opened 2017-03-06
Move report generation to task

KaitCrawford:
Most of this is just moving the functionality over to a task rather than a management command and the tweaks required to facilitate that.


reports/management/commands/generate_reports.py
```diff
@@ -134,601 +69,24 @@ def add_arguments(self, parser):
...
-
-        # Add a warning to the sheet, because we cannot guarantee the best
-        # guess results
-        sheet.add_row({
-            "Timestamp": (
-                "NOTE: This data is not guaranteed to be correct, as "
-                "the current structure of the data does not allow us to link "
-                "the opt out to a subscription or registration, so this is a "
-                "best-effort guess."),
+        generate_report.apply_async(kwargs={
```
rudigiesler: For the management command, do we maybe want it to rather be sync (ie. call the task directly, instead of have it go asyncrously through celery)?

KaitCrawford: That seems reasonable

rudigiesler (approved):

PR #132, opened 2017-03-06
API endpoint for report generation

KaitCrawford:


reports/views.py
```diff
@@ -1 +1,45 @@
...
+            start_date = midnight(timezone.now())
+        if 'end_date' in request.data:
+            end_date = midnight_validator(request.data['end_date'])
+        else:
+            end_date = one_month_after(start_date)
+        email_to = request.data.get('email_to', [])
+        email_from = request.data.get(
+            'email_from', settings.DEFAULT_FROM_EMAIL)
+        email_subject = request.data.get(
+            'email_subject', 'Seed Control Interface Generated Report')
```
rudigiesler: What about creating a serializer for this, and having that do the checking for us? http://www.django-rest-framework.org/api-guide/serializers/

reports/views.py
```diff
@@ -1 +1,45 @@
...
+
+        generate_report.apply_async(output_file=output_file,
+                                    start_date=start_date, end_date=end_date,
+                                    email_recipients=email_to,
+                                    email_sender=email_from,
+                                    email_subject=email_subject)
+        status = 200
```
rudigiesler: I think maybe a 202 response here is more accurate?

reports/views.py
```diff
@@ -1 +1,34 @@
...
+
+    def post(self, request, *args, **kwargs):
+        serializer = ReportGenerationSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        data = serializer.validated_data
+        email_recipients = data.get('email_to', [])
+        email_sender = data.get('email_from', settings.DEFAULT_FROM_EMAIL)
+        email_subject = data.get(
+            'email_subject', 'Seed Control Interface Generated Report')
```
rudigiesler: For these default values, we can use the `default=..` in the serializer.

rudigiesler (approved):

PR #139, opened 2017-06-12
Endpoint for optout from CI

erikh360:


changes/views.py
```diff
@@ -210,3 +210,34 @@ def search_optouts_message_type():
...
+            data = utils.json_decode(request.body)
+        except ValueError as e:
+            return JsonResponse({'reason': "JSON decode error",
+                                'details': six.text_type(e)}, status=400)
+
+        try:
+            identity_id = data['identity']
+        except KeyError as e:
+            return JsonResponse({'reason': '"identity" must be specified.'},
+                                status=400)
```
rudigiesler: Instead of doing all of this validation ourselves, we can just use a serializer: http://www.django-rest-framework.org/api-guide/serializers/

All that we really need to do, is add the `action`, `data`, and `source` fields here, the validation and checking of the request and the `mother_id` field can be done in the serializer.

changes/views.py
```diff
@@ -210,3 +210,34 @@ def search_optouts_message_type():
         'metric_name': total_key,
         'metric_value': total,
     })
+
+
+class ReceiveAdminOptout(mixins.CreateModelMixin,
+                         generics.GenericAPIView):
```
rudigiesler: We're not using the `CreateModelMixin` here, since we're not using the `create` function, which it provides.

We probably also want the `CreateAPIView` as our base class, since that provides the `post` method, which handles validating the request against the serializer, and then creating the object and returning a 201 if everything is okay, else it returns the relevant error message.

rudigiesler (approved):

PR #140, opened 2017-06-14
Add source if it doesn't exist

erikh360:


changes/views.py
```diff
@@ -218,7 +218,15 @@ class ReceiveAdminOptout(generics.CreateAPIView):
...
-        source = Source.objects.get(user=self.request.user)
+        sources = Source.objects.filter(user=self.request.user)
+        if sources:
+            source = sources[0]
+        else:
+            source = Source.objects.create(**{
+                "name": self.request.user.get_full_name(),
+                "user": self.request.user,
+                "authority": "advisor"
+            })
```
rudigiesler: We can probably use `get_or_create` rather than doing this ourselves: https://docs.djangoproject.com/en/1.11/ref/models/querysets/#get-or-create

rudigiesler (approved):

PR #142, opened 2017-06-20
3rd party registration integration

erikh360:
Not done yet, still need to populate the operator_id properly.


registrations/tasks.py
```diff
@@ -390,3 +393,149 @@ def run(
...
+        columns = [
+            "mothers_phone_number",
+            "pregnancy_week",
+            "preferred_msg_language",
+            "type_of_registration",
+            "preferred_msg_type",
+            "message_days",
+            "message_time",
+            "message_receiver",
+            "gatekeeper_phone_number"]
```
rudigiesler: If these are the values in the first row, maybe we should get this list from the document? That way if they switch rows around, it still works?

If not, then this is fine.

registrations/tasks.py
```diff
@@ -390,3 +393,149 @@ def run(
...
+                        utils.get_today(),
+                        line["pregnancy_week"]).strftime("%Y%m%d")
+
+                serializer = RegistrationSerializer(data=reg_info)
+                serializer.is_valid(raise_exception=True)
+
+                Registration.objects.create(**serializer.validated_data)
+            except Exception, error:
+                line['error'] = str(error)
+                ThirdPartyRegistrationError.objects.create(**{'data': line})
```
rudigiesler: I think maybe we should do something like have a variable outside of the loop that keeps the value of the last error, or None if there are no errors. If that variable contains an error after the loop, then we should raise that error, else we do nothing.

That way, if there's an error on any of the rows, we get a Sentry notification about it.

registrations/tasks.py
```diff
@@ -390,3 +393,149 @@ def run(
...
+                            line["pregnancy_week"]).strftime("%Y%m%d")
+                else:
+                    reg_info['data']['baby_dob'] = utils.calc_baby_dob(
+                        utils.get_today(),
+                        line["pregnancy_week"]).strftime("%Y%m%d")
+
+                serializer = RegistrationSerializer(data=reg_info)
+                serializer.is_valid(raise_exception=True)
+
+                Registration.objects.create(**serializer.validated_data)
```
rudigiesler: I think we can use use the `.save` function of the serializer here?

registrations/views.py
```diff
@@ -162,3 +162,17 @@ def get(self, request, *args, **kwargs):
...
+
+class ThirdPartyRegistrationView(APIView):
+
+    """ ThirdPartyRegistrationView Interaction
+        POST - starts up the task that pulls registrations from a 3rd party
+    """
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request, *args, **kwargs):
+        status = 200
```
rudigiesler: Maybe a better code is `201`? Since we're telling the client that we created the task?

registrations/tests.py
```diff
@@ -3314,3 +3314,101 @@ def test_generate_and_send_invalid_json(
...
+        "message_time": "2-5pm",
+        "message_receiver": "",
+        "gatekeeper_phone_number": ""
+    }]
+
+
+class TestThirdPartyRegistrations(AuthenticatedAPITestCase):
+
+    def setUp(self):
+        tasks.PullThirdPartyRegistrations.get_data = override_get_data
```
rudigiesler: This is quite a large function, that does a lot. I think we should have tests that cover it. I don't think that these existing tests need to cover the `get_data` function, but I think there should be additional tests covering the functionality of the `get_data` function.

registrations/tasks.py
```diff
@@ -390,3 +393,164 @@ def run(
...
+            "mother_and_family": "mother_family",
+            "mother_and_friend": "mother_friend",
+        }.get(receiver, receiver)
+
+    def get_data(self):
+        url = settings.THIRDPARTY_REGISTRATIONS_URL
+        username = settings.THIRDPARTY_REGISTRATIONS_USER
+        password = settings.THIRDPARTY_REGISTRATIONS_PASSWORD
+
+        response = requests.get(url, verify=False, auth=(username, password))
```
rudigiesler: Is there a reason why we're not verifying the SSL certificate here? If they have issues, I think we should error on that. This is sensitive information, we should make sure that the connection is secure.

If they have issues with their certificates, we need to tell them to fix it.

registrations/tasks.py
```diff
@@ -390,3 +393,164 @@ def run(
...
+                        "msg_receiver": receiver,
+                        "operator_id": operator_id,
+                        "language": language,
+                        "msg_type": msg_type
+                    }
+                }
+
+                if line['gatekeeper_phone_number']:
+                    identity_details["receiver_role"] = receiver.replace(
+                        'mother_', '')
```
rudigiesler: I'm a little confused by this. Why are we changing the receiver type if there's a gatekeeper phone number. Surely if they want `mother_and_father` messages, then we should be sending to both the mother and the father? Not just the father?

registrations/tasks.py
```diff
@@ -390,3 +393,164 @@ def run(
...
+
+    def get_voice_days(self, days):
+        return {'monday_and_wednesday': 'mon_wed',
+                'tuesday_and_thursday': 'tue_thu'}.get(days, days)
+
+    def get_receiver(self, receiver):
+        return {
+            "mother_and_father": "mother_father",
+            "mother_and_family": "mother_family",
+            "mother_and_friend": "mother_friend",
```
rudigiesler, line 450: What about the other receiver types? The `_only` receiver types.

registrations/tasks.py
erikh360: I got this from the jsbox app but I might have misunderstood something, I will relook at this.

erikh360: This is only the receiver role of the relevant identity as per:
https://github.com/praekelt/mama-ng-jsbox/blob/develop/go-voice_registration.js#L680
I'm reshuffling some things to update the identity details correctly, commit to follow soon.

registrations/tasks.py
erikh360: It's only these three where the string doesn't match the string on our side, I use a get with the receiver as default.

rudigiesler (approved):

PR #143, opened 2017-06-22
Add generic change endpoint

erikh360:
This endpoint will create a change from external systems for updating language and messageset.

### DIFF
--- a/changes/serializers.py
+++ b/changes/serializers.py
@@ -33,6 +33,7 @@ class ChangeSerializer(serializers.ModelSerializer):
 
 class AdminChangeSerializer(serializers.Serializer):
     mother_id = serializers.UUIDField(allow_null=False)
+    subscription = serializers.UUIDField(required=False)
     messageset = serializers.CharField(required=False)
     language = serializers.CharField(required=False)
 
--- a/changes/serializers.py
+++ b/changes/serializers.py
@@ -2,6 +2,25 @@ from .models import Change
 from rest_framework import serializers
 
 
+class OneFieldRequiredValidator:
+    def __init__(self, fields):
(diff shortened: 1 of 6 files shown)


KaitCrawford (commented):
This is probably a bit pedantic but I think either the view or serialiser should check that the language chosen is a valid one for the messageset chosen.
Sorry this fell off my radar last week.

erikh360:
@KaitCrawford no worries, let me add the check.

erikh360:
@KaitCrawford ready for another check.

changes/serializers.py
```diff
@@ -21,6 +25,39 @@ def __call__(self, data):
...
+
+            if data.get('language'):
+                new_lang = data['language']
+
+                sbmApi = StageBasedMessagingApiClient(
+                    api_url=settings.STAGE_BASED_MESSAGING_URL,
+                    auth_token=settings.STAGE_BASED_MESSAGING_TOKEN
+                )
+
+                messagesets = []
```
KaitCrawford, line 44: Why is this a list? Can there be more than one messageset or subscription in the change?

KaitCrawford (commented):
Looks good. Just one comment

changes/serializers.py
erikh360: Yes, the `change_language` task updates all subscriptions linked https://github.com/praekelt/hellomama-registration/blob/develop/changes/tasks.py#L197

KaitCrawford: ah, I'd missed the `for subscription in subscriptions:` loop further down

KaitCrawford (approved):
:+1:

PR #146, opened 2017-06-26
Change to SEED metrics client

rudigiesler:
Currently, we are using the vumi go metrics client. While this works with the seed metrics API, it uses token authentication, rather than the basic authentication that the seed metrics API uses. So in order to use authentication for our metrics, we need to change to the seed metrics client.


KaitCrawford (approved):
:+1: when travis

PR #147, opened 2017-07-10
Fix new_rate var not declared

erikh360:


changes/tests.py
```diff
@@ -547,6 +547,100 @@ def test_prebirth_text_to_audio_week28_new_short_name(self):
...
+            responses.GET,
+            'http://localhost:8005/api/v1/messageset/%s' % query_string,
+            json={
+                "count": 1,
+                "next": None,
+                "previous": None,
+                "results": [{
+                    "id": 4,
+                    "short_name": 'prebirth.mother.text.0_9',
+                    "default_schedule": 1
```
KaitCrawford, line 624: Shouldn't the new subscription and the old subscription be on different schedules if you want to test the new `next_sequence_number`?

KaitCrawford (commented):
Left a comment

changes/tests.py
erikh360: I'm simulating a text to text change which is on the same schedule, the test above this one changes to a different schedule.

KaitCrawford: ah ok

KaitCrawford (approved):
:+1:

PR #148, opened 2017-07-18
[SEED-909] Get identity using seed-services-client

alexmuller:
There are many similar cases in this file and across this project. This is the smallest possible change I can make to be sure I'm doing the right thing.

Use the API client to send requests to avoid duplicating code to talk to the seed-message-sender API.

### DIFF
--- a/hellomama_registration/utils.py
+++ b/hellomama_registration/utils.py
@@ -6,12 +6,19 @@ import six
 from django.conf import settings
 from registrations.models import Source
 from datetime import timedelta
+from seed_services_client import IdentityStoreApiClient
 
 session = requests.Session()
 http = requests.adapters.HTTPAdapter(max_retries=5)
 https = requests.adapters.HTTPAdapter(max_retries=5)
 session.mount('http://', http)
 session.mount('https://', https)
+
+identity_store_client = IdentityStoreApiClient(
+    api_url=settings.IDENTITY_STORE_URL,
+    auth_token=settings.IDENTITY_STORE_TOKEN,
+    retries=5,
+)
 
 
 def get_today():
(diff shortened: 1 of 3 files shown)


hellomama_registration/utils.py
```diff
@@ -6,6 +6,7 @@
 from django.conf import settings
 from registrations.models import Source
 from datetime import timedelta
+from seed_services_client import MessageSenderApiClient
 
 session = requests.Session()
 http = requests.adapters.HTTPAdapter(max_retries=5)
```
erikh360: We add retries here

hellomama_registration/utils.py
```diff
@@ -301,15 +302,9 @@ def get_messageset_schedule_sequence(short_name, weeks):
...
-    result = session.post(
-        url="%s/outbound/" % settings.MESSAGE_SENDER_URL,
-        data=json.dumps(payload),
-        headers={
-            'Content-Type': 'application/json',
-            'Authorization': 'Token %s' % (
-                settings.MESSAGE_SENDER_TOKEN,)
-        }
-    ).json()
+    ms_client = MessageSenderApiClient(settings.MESSAGE_SENDER_TOKEN,
```
erikh360: We could send the session here.

erikh360 (commented):
This looks good but it looks like we would be losing the  retry functionality. We could send the session to the client or update the client to always retry.

hellomama_registration/utils.py
alexmuller: Ah thanks! I've passed in the session. I think it'd probably be nicer for the API client to handle the retrying logic with a simple `retry` (or whatever) parameter but we I guess we can sort that out another time.

alexmuller: That's exciting, passing in the session causes 15 tests to fail. Will investigate why!

erikh360: I think it's a different session type or something, so passing in the session is maybe not a proper solution. We could pass in the correct one or just handle it properly in the client.

alexmuller: praekeltfoundation/seed-services-client#33

alexmuller:
Ok cool, I've changed this slightly to use an example of `IdentityStoreApiClient` to make sure I'm doing the right thing. I think this is ready for another review!

erikh360 (approved):
Looks good to me.

PR #149, opened 2017-07-26
Add new registration endpoint

erikh360:


registrations/views.py
```diff
@@ -177,3 +177,30 @@ def post(self, request, *args, **kwargs):
...
+    """ ThirdPartyRegistrationView Interaction
+        POST - Validates and Saves the registration
+    """
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request, *args, **kwargs):
+        status = 201
+        resp = {"registration_added": True}
+
+        source = Source.objects.get(user=self.request.user.id)
```
rudigiesler: Maybe we should have some handling in case there's no source for that user?

rudigiesler (commented):
Just the one comment from me

registrations/views.py
erikh360: I opted for not creating the `source` automatically like we do in other places, we might not know the `authority` of the `source` in the code so it would have to be created manually.

rudigiesler: :+1: , I was just thinking about returning a proper error back instead of a 500, not automatically adding a source.

rudigiesler (approved):

PR #151, opened 2017-07-28
Enabled patch endpoint on registration

erikh360:


registrations/views.py
```diff
@@ -104,17 +104,22 @@ def post(self, request, *args, **kwargs):
...
     serializer_class = RegistrationSerializer
+    lookup_field = 'id'
 
     def post(self, request, *args, **kwargs):
         # load the users sources - posting users should only have one source
         source = Source.objects.get(user=self.request.user)
         request.data["source"] = source.id
         return self.create(request, *args, **kwargs)
 
+    def patch(self, request, *args, **kwargs):
```
KaitCrawford, line 120: is this method necessary? doesn't the update mixin already provide this?

KaitCrawford (commented):
looks good. 1 small query though

registrations/views.py
erikh360: @KaitCrawford I get a `405 Method Not Allowed` error if I remove it.

KaitCrawford: odd. but ok

KaitCrawford (approved):
:+1:

registrations/views.py
rudigiesler: The `UpdateModelMixin` gives us the `partial_update` and `update` methods. The `UpdateApiView` class is the one that provides the `put` and `patch` classes that uses these mixins: https://github.com/encode/django-rest-framework/blob/a251b9379200420062cad9e3c68fde7c0e6b3fdc/rest_framework/generics.py#L222-L231

KaitCrawford: Ah! thanks for the clarification Rudi

PR #153, opened 2017-08-01
Make report xlsx filename random

erikh360:


reports/tests.py
```diff
@@ -90,6 +96,13 @@ def has_listeners(class_name):
         post_save.connect(receiver=model_saved,
                           dispatch_uid='instance-saved-hook')
 
+        try:
+            with mock.patch('random.choice', mockobj.choice):
+                filename = generate_random_filename()
+                os.remove(filename)
```
KaitCrawford, line 102: I'm confused about why we need this? You're mocking the `choice` function, creating a new file and then removing it right?

reports/tests.py
```diff
@@ -314,21 +322,22 @@ def midnight(self, timestamp):
...
+                                                            '%Y-%m-%d')),
+                'email_recipients': ['foo@example.com'],
+                'email_subject': 'The Email Subject'})
 
-        return tmp_file
+            return filename
 
     @responses.activate
-    def test_generate_report_email(self):
+    @mock.patch("os.remove")
```
KaitCrawford, line 339: if we're mocking the removal in all these tests won't we end up with a bunch of files that aren't deleted at the end of the tests?

KaitCrawford (commented):
I've left some questions about the tests

reports/tests.py
erikh360: This is to actually delete the file, I mock choice to get the same filename as the tests. It only generates a name, not the actual file.

reports/tests.py
erikh360: I remove the file in tearDown, I mock it in all tests because we want to check the contents before it gets deleted.

reports/tests.py
KaitCrawford: aaah ok! I see now! thanks :)

KaitCrawford (approved):
:+1: thanks for the clarifications

PR #154, opened 2017-08-02
Fix for empty to_addr field

erikh360:
Recently we deployed changes for POPI compliance which meant we removed the msisdn from the `to_addr` field on all outbounds in the message sender database.

The CI report sms delivery tab grouped by the `to_addr` field, so all messages were added to one single line where the msisdn was empty, the result was more columns than excel can handle and it failed.

This PR will do a identity lookup if the msisdn is not present.

### DIFF
--- a/reports/tasks.py
+++ b/reports/tasks.py
@@ -98,6 +98,7 @@ class GenerateReport(Task):
 
         self.identity_cache = {}
         self.messageset_cache = {}
+        self.address_cache = {}
 
         ids_client = IdentityStoreApiClient(settings.IDENTITY_STORE_TOKEN,
                                             settings.IDENTITY_STORE_URL)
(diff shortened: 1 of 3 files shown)


reports/tasks.py
```diff
@@ -364,6 +365,19 @@ def handle_sms_delivery_msisdn(
...
+                    identity = self.get_identity(
+                        ids_client, outbound['to_identity'])
+
+                    details = identity.get('details', {})
+                    default_addr_type = details.get(
+                        'default_addr_type', 'msisdn')
+
+                    addresses = details.get('addresses', {})
+                    outbound['to_addr'] = list(
+                        addresses.get(default_addr_type, {}).keys())[0]
```
KaitCrawford: some msisdns have the `default` key so we should probably prefer those over others

KaitCrawford (commented):
Looks good. 1 small comment

reports/tasks.py
rudigiesler: This was recently added to the seed services client: https://github.com/praekeltfoundation/seed-services-client/blob/develop/seed_services_client/identity_store.py#L44-L54 , maybe we can use that instead?

erikh360:
@KaitCrawford Ready for another look!

alexmuller:
Is it https://github.com/praekelt/seed-message-sender/pull/52 which is causing the blank `to_addr`?

I think I understand what's going on with this change 👍

erikh360:
@alexmuller Yes it was #52, we replace the msisdn with a identity ID to make it easier to delete a user's details when they opt out. 

Happy for me to merge?

alexmuller (approved):
Makes sense to me 👍

PR #156, opened 2017-08-04
Add support for household

erikh360:


alexmuller:
I think this makes sense assuming that the POST body in the test is correct 👍, but somebody else apart from me should probably have a look too.

My understanding:

- We used to only get unsubscribe requests for the mother
- Now we can get unsubscribe requests for the household but it's still part of the mother's MSISDN
- `'234'` is the country code for Nigeria (it's a magic number in the diff here)

Maybe outside the scope of this PR but the `normalize_msisdn` function in this repo probably doesn't need to take a `country_code` because everything in here is related to Nigeria. But I understand [the `normalize_msisdn` function has a long history](https://github.com/search?q=org%3Apraekelt+normalize_msisdn&type=Code&utf8=%E2%9C%93) :)

erikh360:
@alexmuller This whole endpoint is new, the jsbox apps created changes with uuids in mother_id and household_id, this endpoint is for a 3rd party that doesn't have access to our db so they send msisdns and I convert it to uuids. 

Then I create a change object as per normal and a task kicks off that handles it. 

My previous PR (https://github.com/praekelt/hellomama-registration/pull/155) covered most of the functionality but I forgot the household id lookup.

KaitCrawford (approved):
:+1:

PR #158, opened 2017-08-08
Standardize change endpoint

erikh360:
I changed this to accept the same values as the `addregistration` endpoint, it is the same values as we are getting from a 3rd party.


KaitCrawford (approved):
:+1:

PR #160, opened 2017-08-16
CI Report task visibility

erikh360:
The report that is run from the CI fires of the task without much feedback to the user, the user would wait for a email that never arrives not knowing anything about what happened.

This will give the user some visibility on where in the process the task is and if they should wait for the email or escalate/retry.

The plan is to show the last 10 tasks below the report form on the CI, the user can then see what is going on.


reports/tasks.py
```diff
@@ -84,67 +88,80 @@ class GenerateReport(Task):
...
+                    'file_name': file_name,
+                    'file_location': output_file,
+                    'sender': email_sender,
+                    'recipients': email_recipients,
+                    "task_status": task_status})
+        except Exception as error:
+            task_status.status = 'Failed'
+            task_status.error = str(error)
+            task_status.save()
+            raise
```
rudigiesler: Instead of wrapping everything in a try/catch, we can use celery's `on_failure` method. eg. https://github.com/praekelt/seed-stage-based-messaging/blob/develop/subscriptions/tasks.py#L309-L318

reports/views.py
```diff
@@ -19,12 +22,31 @@ def post(self, request, *args, **kwargs):
...
+            "start_date": datetime.strftime(data['start_date'], '%Y-%m-%d'),
+            "end_date": datetime.strftime(data['end_date'], '%Y-%m-%d'),
+            "email_subject": data['email_subject'],
+            "status": "Pending"
+        })
+
         generate_report.apply_async(kwargs={
             "start_date": datetime.strftime(data['start_date'], '%Y-%m-%d'),
             "end_date": datetime.strftime(data['end_date'], '%Y-%m-%d'),
+            "task_status": task_status,
```
rudigiesler: You usually don't want to send whole django objects to the task. What you want to do is send the primary key of the model, and load it when the task loads. This ensures that "asserting the world is the responsibility of the task" http://docs.celeryproject.org/en/latest/userguide/tasks.html#state

reports/models.py
```diff
@@ -1 +1,21 @@
...
+
+@python_2_unicode_compatible
+class ReportTaskStatus(models.Model):
+    """ The status of the report generate tasks
+    """
+
+    start_date = models.CharField(max_length=30, null=False, blank=False)
+    end_date = models.CharField(max_length=30, null=False, blank=False)
+    email_subject = models.CharField(max_length=100, null=False, blank=False)
+    file_size = models.CharField(max_length=30, null=True)
```
rudigiesler: Why don't we just store the exact value in bytes here, and dynamically generate the user friendly string when we need it. We can probably have that as a function on the model, the django convention is usually `get_XXX_display`, ie `get_file_size_display`

reports/models.py
```diff
@@ -1 +1,21 @@
...
+    status = models.CharField(max_length=30, null=False, blank=False)
```
rudigiesler: We probably want this to have a `choices` attribute, with a limited amount of choices that you can select between. This way we can tell all the possible values for this field by looking at the model description.

reports/models.py
```diff
@@ -1 +1,21 @@
...
+    error = models.CharField(max_length=255, null=True)
```
rudigiesler: Since errors have arbitrary length, maybe a `TextField` might be better here?

reports/views.py
```diff
@@ -19,12 +22,31 @@ def post(self, request, *args, **kwargs):
...
         resp = {"report_generation_requested": True}
         return Response(resp, status=status)
+
+
+class ReportTaskStatusViewSet(viewsets.ReadOnlyModelViewSet):
+    """
+    API endpoint that allows Registrations to be viewed.
+    """
+    permission_classes = (IsAuthenticated,)
+    queryset = ReportTaskStatus.objects.all().order_by('-created_at')[:10]
```
rudigiesler: Instead of only showing the last 10 results, what about paginating the result, with a page size of 10? That way we can view older tasks if we want to.

reports/utils.py
```diff
@@ -37,3 +37,11 @@ def midnight_validator(inputstr):
...
     return ''.join(
         random.choice(string.ascii_lowercase) for i in range(12)) + suffix
+
+
+def sizeof_format(num, suffix='B'):
+    for unit in ['', 'Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi']:
+        if abs(num) < 1024.0:
+            return "%3.1f%s%s" % (num, unit, suffix)
+        num /= 1024.0
+    return "%.1f%s%s" % (num, 'Yi', suffix)
```
rudigiesler: We've got a function in Django to do this for us: https://docs.djangoproject.com/en/dev/ref/templates/builtins/#filesizeformat

reports/models.py
erikh360: Is it better to then maybe save the integer and format it where we use it, the CI, the `get_XXX_display` doesn't get called from the view.

rudigiesler: I think in either place is fine. To get it on this view, you just need to add a `get_file_size` function to your serializer.

erikh360:
This is ready for another look @rudigiesler

reports/models.py
```diff
@@ -1 +1,29 @@
...
+@python_2_unicode_compatible
+class ReportTaskStatus(models.Model):
+    """ The status of the report generate tasks
+    """
+
+    STATUS_CHOICES = (
+        ('Pending', "Pending"),
+        ('Running', "Running"),
+        ('Sending', "Sending"),
+        ('Done', "Done")
```
rudigiesler: Usually for this case, we do something like
```
PENDING = 'P'
RUNNING = 'R'
....
STATUS_CHOICES = (
    (PENDING, 'Pending'),
...
)
```
This allows us to do three things. We can reference the value using `ReportTaskStatus.PENDING`, so we can immediately see whether we've set it to a valid value or not (Editor can autocomplete `status = ReportTaskStatus.PENDING`, but it cannot autocomplete `status = 'Pending'`), it allows us to use shorter values like `P` instead of `Pending`, but in our code we still know exactly what we mean, but take up less storage on the database server, and it ensures that we don't make mistakes like `if status == 'pending'`, which would be an easy mistake to make, and would result in the content of the if statement never being able to run.

Docs: https://docs.djangoproject.com/en/1.11/ref/models/fields/#choices

rudigiesler (commented):
Just one last thing from me, looks good otherwise :)

reports/models.py
erikh360: Thanks @rudigiesler I just copied a choice field from the registration models, it's better now and I also realized there was a missing choice with this implementation.

Ready for another look!

rudigiesler (approved):

rudigiesler (approved):

PR #161, opened 2017-08-21
Use cursor pagination for api endpoints

KaitCrawford:


registrations/management/commands/fix_registration_subscriptions.py
```diff
@@ -70,7 +70,7 @@ def handle(self, *args, **kwargs):
             sub_response = client.get_subscriptions({
                 'identity': receiver_id,
             })
-            if sub_response['count']:
+            if len(sub_response['results']):
```
rudigiesler: Do we not need to do `len(list(...))` here instead?

rudigiesler (commented):
Just one comment from me

alexmuller (approved):

PR #162, opened 2017-08-22
Fix household only and identity patch

erikh360:
Fix for changes household identities - there is no mother msisdn to send so we need to get it from the father/friend/family identity's `linked_to` field.

The identity patch was incorrect, also fixed that.

### DIFF
--- a/changes/tests.py
+++ b/changes/tests.py
@@ -3438,7 +3438,7 @@ class AddChangeViewsTest(AuthenticatedAPITestCase):
     """
     Tests related to the adding of changes view.
     """
-    def mock_identity_lookup(self, msisdn, identity_id):
+    def mock_identity_lookup(self, msisdn, identity_id, details={}):
         responses.add(
             responses.GET,
             'http://localhost:8001/api/v1/identities/search/?details__addresses__msisdn=%s' % msisdn,  # noqa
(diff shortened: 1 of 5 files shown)


changes/tests.py
```diff
@@ -3593,10 +3593,50 @@ def test_add_change_unsubscribe_household(self):
...
+                                         json.dumps(post_data),
+                                         content_type='application/json')
+
+        # Check
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
+
+        d = Change.objects.last()
+        self.assertEqual(d.source.name, 'test_ussd_source_adminuser')
+        self.assertEqual(d.action, 'unsubscribe_household_only')
+        self.assertEqual(d.validated, False)
```
rudigiesler, line 3627: Should the task not be validated? Or is this with the validation task disabled? If so, do we have test checking that a Change with this type of data will validate successfully?

erikh360: There is no code that validates changes at the moment. All changes in Production are validated=False.

I have spoken to Chima about this and there should be a ticket for validation soon.

rudigiesler: Oh right, so we're just keeping track of the changes, not doing anything with them?

In that case, :+1: on this PR

erikh360: There is a task that fires and processes the change, but it doesn't validate the data and set the flag like MC.

PR #163, opened 2017-08-23
Gatekeeper should be linked to mother

erikh360:
The `linked_to` field on the gatekeeper(father/friend/family) should be populated with the mother's identity id.

### DIFF
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -525,10 +525,15 @@ class PullThirdPartyRegistrations(Task):
 
             if receiver in ('father_only', 'family_only', 'friend_only'):
                 mother_identity = self.get_or_create_identity(
-                    None, receiver_identity['id'])
+                    None, receiver_identity['id'], identity_details)
                 reg_info['mother_id'] = mother_identity['id']
 
-        if mother_identity:
+                receiver_details["linked_to"] = mother_identity['id']
+                utils.patch_identity(receiver_identity['id'],
+                                     {'details': receiver_details})
+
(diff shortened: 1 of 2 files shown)


registrations/tasks.py
```diff
@@ -525,10 +525,15 @@ def create_registration(self, line, source):
...
-                    None, receiver_identity['id'])
+                    None, receiver_identity['id'], identity_details)
                 reg_info['mother_id'] = mother_identity['id']
 
-        if mother_identity:
+                receiver_details["linked_to"] = mother_identity['id']
+                utils.patch_identity(receiver_identity['id'],
+                                     {'details': receiver_details})
+
+        if (mother_identity and
```
KaitCrawford, line 535: Is this necessary? Won't `mother_identity` always be populated since it's created at the start of the method? (L469)

KaitCrawford (commented):
One small question

registrations/tasks.py
erikh360: L469 only creates a identity if we receive a mother msisdn, in household only cases we don't.

This patches the `receiver_identity`(father/friend/family) with the `linked_to` field populated with the mother id we just created on L572. It's complicated because we have to create the receiver to get the id, then create a mother linked to the receiver, then link the receiver back to the mother. It was not fun figuring this out in the jsbox apps.

KaitCrawford: ah I see.

KaitCrawford (approved):
:+1:

PR #164, opened 2017-08-24
Fix address being removed

erikh360:
I used the original dictionary instead of the results from the identity store client to patch the identity, it doesn't include the address so it was removed, this fixes that.


KaitCrawford (approved):
:+1:

PR #165, opened 2017-08-24
[SEED-909] Use seed-services-client for utils.py

alexmuller:
There are 2 uses here which involve pagination which I think will be changed as part of @KaitCrawford's pagination work so I've marked them with `FIXME`s for now.

### DIFF
--- a/hellomama_registration/utils.py
+++ b/hellomama_registration/utils.py
@@ -103,10 +103,14 @@ def get_identity_address(identity):
         return None
 
 
-def search_identities(params):
-    """ Returns the identities matching the given parameters
+def search_identities(search_key, search_value):
+    """
+    Returns the identities matching the given parameters
+    FIXME: This should be handled by identity_store_client when
+    it supports pagination
     """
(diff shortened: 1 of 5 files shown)


hellomama_registration/utils.py
```diff
@@ -89,24 +105,17 @@ def get_identity(identity):
...
-    }
-    r = session.get(url, params=params, headers=headers).json()
-    if len(r["results"]) > 0:
-        return r["results"][0]["address"]
-    else:
-        return None
+    return identity_store_client.get_identity_address(identity)
 
 
-def search_identities(params):
```
KaitCrawford, line 106: Why the change to the arguments here?

alexmuller: Ah sorry I put it in the commit message but didn't copy it into the pull request:

This behaviour more closely matches the seed-services API client so it will make it easier to use the API client when that supports pagination.

KaitCrawford (approved):
:+1:

PR #166, opened 2017-08-25
Add personnel codes API endpoint

erikh360:
This is for vas2nets to update their list of personnel codes to validate against. It's in this repo because it is hellomama specific and I don't want them to call the Identity-store directly because there might be personal info on the identity that we don't want to share.

### DIFF
--- a/registrations/tests.py
+++ b/registrations/tests.py
@@ -3960,9 +3960,7 @@ class TestPersonnelCodeView(AuthenticatedAPITestCase):
                                          content_type='application/json')
 
         self.assertEqual(response.status_code, 200)
-
-        results = json.loads(response.content)['results']
-        self.assertEqual(results, ['11111', '22222'])
+        self.assertEqual(response.data['results'], ['11111', '22222'])
 
     @responses.activate
     def test_return_personnel_codes_post(self):
--- a/registrations/tests.py
+++ b/registrations/tests.py
@@ -3960,4 +3960,4 @@ class TestPersonnelCodeView(AuthenticatedAPITestCase):
                                          content_type='application/json')
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.data['results'], ['11111', '22222'])
+        self.assertEqual(sorted(response.data['results']), ['11111', '22222'])
--- a/registrations/tests.py
+++ b/registrations/tests.py
@@ -3924,3 +3924,64 @@ class TestAddRegistrationsAPI(TestThirdPartyRegistrations):
             response.data['error'].find("This field may not be null") != -1)
 
         self.assertEqual(Registration.objects.count(), 0)
+
+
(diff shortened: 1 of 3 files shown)


registrations/views.py
```diff
@@ -213,3 +213,23 @@ def post(self, request, *args, **kwargs):
...
+
+    def get(self, request, *args, **kwargs):
+        status = 200
+        resp = {"results": []}
+
+        identities = utils.search_identities(
+            {"details__has_key": "personnel_code"})
+
+        for identity in identities:
+            resp["results"].append(identity['details'].get('personnel_code'))
```
rudigiesler: I think that we should maybe put these codes in a set rather, to avoid duplicates.

registrations/views.py
```diff
@@ -213,3 +213,23 @@ def post(self, request, *args, **kwargs):
             status = 400
 
         return Response(resp, status=status)
+
+
+class PersonnelCodeView(APIView):
+
+    """ PersonnelCodeView Interaction
+        GET - returns a list of identities with personnel codes
```
rudigiesler: This doesn't seem to be correct. It returns a list of personnel codes, not identities.

registrations/views.py
```diff
@@ -213,3 +213,23 @@ def post(self, request, *args, **kwargs):
...
+        status = 200
+        resp = {"results": []}
+
+        identities = utils.search_identities(
+            {"details__has_key": "personnel_code"})
+
+        for identity in identities:
+            resp["results"].append(identity['details'].get('personnel_code'))
+
+        return Response(resp, status=status)
```
rudigiesler: What are we trying to achieve with this view?

If the goal is to have an endpoint to check whether a clinic code is valid or not, then maybe we should do that, instead of returning all valid clinic codes.

If the client of this endpoint is going to use all of the clinic codes, then this endpoint is fine.

registrations/tests.py
```diff
@@ -3924,3 +3924,62 @@ def test_add_registration_invalid_data(self):
...
+    @responses.activate
+    def test_return_personnel_codes_patch(self):
+        """
+        It should not allow PATCH method
+        """
+
+        response = self.normalclient.patch('/api/v1/personnelcode/', {},
+                                           content_type='application/json')
+
+        self.assertEqual(response.status_code, 405)
```
rudigiesler: I'm not sure that the POST and PATCH tests are entirely necessary. It almost feels like we're testing django rest framework here rather than our own code.

registrations/views.py
erikh360: The goal of the vas2nets line is to work even when mainone is down, so they would request this periodically and save it to validate against. If mainone is down then they validate with what they have saved.

rudigiesler (approved):

rudigiesler (approved):

PR #167, opened 2017-09-13
Upgrade seed services client to 0.31.0

KaitCrawford:


reports/tasks.py
```diff
@@ -205,37 +204,16 @@ def get_registrations(self, **kwargs):
...
         subscriptions = sbm_client.get_subscriptions(kwargs)
-        cursor = subscriptions['next']
-        while cursor:
-            for result in subscriptions['results']:
-                yield result
-            params = parse_cursor_params(cursor)
-            subscriptions = sbm_client.get_subscriptions(params)
-            cursor = subscriptions['next']
         for result in subscriptions['results']:
             yield result
```
rudigiesler: Can we not just do a `return sbm_client.get_subscriptions(kwargs)['results']` here?

rudigiesler (commented):
Just the one comment from me

reports/tasks.py
KaitCrawford: I removed these methods altogether and just called the client directly

changes/tasks.py
```diff
@@ -128,6 +128,7 @@ def change_messaging(self, change):
         current_rate = len(current_days.split(','))  # msgs per week
 
         # Deactivate subscriptions
+        utils.deactivate_subscription(current_sub)
```
alexmuller, line 131: I don't really understand this addition. I mean it seems perfectly correct but I don't understand how it was working before? :)

alexmuller: Ah sorry I do understand now. Because we're pulling the first element of the generator we need to process that one separately to the ones below 👍

alexmuller (approved):
Nice, looks tidier!

rudigiesler (approved):

PR #168, opened 2017-09-26
Transfer registrations between facilities

erikh360:


registrations/management/commands/transfer_registration_facility.py
```diff
@@ -0,0 +1,58 @@
...
+
+        for identity in utils.search_identities("details__personnel_code",
+                                                from_code):
+            from_identity = identity['id']
+            break
+
+        for identity in utils.search_identities("details__personnel_code",
+                                                to_code):
+            to_identity = identity['id']
+            break
```
rudigiesler: I think we'll have an issue here if we don't find any identities with that personnel code?

Maybe it's better if we check for `next(..)`, and catch the `StopIteration` and raise a `CommandError` instead?

registrations/management/commands/transfer_registration_facility.py
```diff
@@ -0,0 +1,58 @@
...
+            registration.save()
+            updated += 1
+
+        self.success('Updated %s registrations.' % updated)
+
+    def log(self, level, msg):
+        self.stdout.write(level(msg))
+
+    def warning(self, msg):
+        self.log(self.style.WARNING, msg)
```
rudigiesler: I don't think that this function is being used anywhere?

registrations/tests.py
```diff
@@ -4050,3 +4050,77 @@ def test_return_personnel_codes(self):
 
         self.assertEqual(response.status_code, 200)
         self.assertEqual(sorted(response.data['results']), ['11111', '22222'])
+
+
+class TestTransferRegistrationsCommand(AuthenticatedAPITestCase):
+
+    def make_registration(self, operator_id):
+        reg_data = REG_DATA['hw_pre_mother']
```
rudigiesler: Maybe we should make a copy of this dictionary instead? So that when we modify it below, we're not changing it for future tests.

registrations/management/commands/transfer_registration_facility.py
rudigiesler: Also, do we have any guarantee that there will only be one identity for each personnel code? Maybe to just be safe, we should iterate through all of the `from_identity`s that we find, and do the below step for each identity, to make sure that we catch all of the registrations for that code?

rudigiesler (approved):

PR #169, opened 2017-09-27
Pull vas2nets call data

erikh360:


vas2nets/models.py
```diff
@@ -0,0 +1,13 @@
+from django.utils.encoding import python_2_unicode_compatible
+from django.db import models
+
+
+@python_2_unicode_compatible
+class VoiceCall(models.Model):
+    created_at = models.DateTimeField()
+    msisdn = models.CharField(null=False, blank=False, max_length=36)
+    duration = models.IntegerField(null=False, blank=False)
+    reason = models.CharField(null=False, blank=False, max_length=36)
```
rudigiesler: What about having a `choices` attribute on here, since the reason field is limited to a specific set of choices?

vas2nets/views.py
```diff
@@ -0,0 +1,23 @@
...
+class FetchVoiceDataView(APIView):
+
+    """ FetchVoiceData Interaction
+        POST - starts up the task that pulls voice data from the vas2nets
+        portal
+    """
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request, *args, **kwargs):
+        status = 201
```
rudigiesler: Maybe 202 Accepted is a better response code here?

vas2nets/views.py
```diff
@@ -0,0 +1,23 @@
...
+        date = utils.get_today() - timedelta(1)
+        fetch_voice_data.apply_async(args=[date.strftime('%Y-%m-%d')])
+        resp = {"fetch_voice_data_initiated": True}
```
rudigiesler: Maybe we can return the UUID of the task in the response? In case we ever want to keep track of it?

vas2nets/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+
+        content = requests.get(url).content.decode('utf-8')
+
+        return list(csv.reader(content.splitlines(), delimiter=','))[1:]
+
+    def run(self, date, **kwargs):
+        data = self.get_data(date)
+        for row in data:
+            VoiceCall.objects.create(
+                created_at=row[0],
```
rudigiesler: We probably want to add timezone data to this datetime.

Can we figure out what timezone they're sending in the CSV, and add that data to the datetime object before storing it in the DB?

vas2nets/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class FetchVoiceData(Task):
+
+    def get_data(self, date):
+        url = "%s?report_date=%s" % (settings.V2N_VOICE_URL, date)
+
+        content = requests.get(url).content.decode('utf-8')
+
+        return list(csv.reader(content.splitlines(), delimiter=','))[1:]
```
rudigiesler: Maybe we can use a DictReader here instead? That way if they reorder/add columns, this will still work fine.

vas2nets/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def run(self, date, **kwargs):
+        data = self.get_data(date)
+        for row in data:
+            VoiceCall.objects.create(
+                created_at=row[0],
+                msisdn=row[2],
```
rudigiesler: What about the shortcode field that they give us? We can store that fairly easily, and we might want it for future queries, especially if we want to compare stats between the 55500 and the 1444 line.

vas2nets/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+
+from .models import VoiceCall
+
+
+class FetchVoiceData(Task):
+
+    def get_data(self, date):
+        url = "%s?report_date=%s" % (settings.V2N_VOICE_URL, date)
+
+        content = requests.get(url).content.decode('utf-8')
```
rudigiesler: Maybe we should do a `stream=True` in the request, and then do a `.iter_lines()`, because it does seem like they're streaming the data on their side, and it will allow us to not have to load all the data into memory before processing it.

http://docs.python-requests.org/en/master/user/advanced/#streaming-requests

vas2nets/tests.py
```diff
@@ -0,0 +1,84 @@
...
+
+class AuthenticatedAPITestCase(APITestCase):
+
+    def make_source_normaluser(self):
+        data = {
+            "name": "test_voice_source_normaluser",
+            "authority": "patient",
+            "user": User.objects.get(username='testnormaluser')
+        }
+        return Source.objects.create(**data)
```
rudigiesler: It doesn't seem like we're using this function anywhere?

vas2nets/tasks.py
```diff
@@ -0,0 +1,29 @@
...
+
+        return list(csv.reader(content.splitlines(), delimiter=','))[1:]
+
+    def run(self, date, **kwargs):
+        data = self.get_data(date)
+        for row in data:
+            VoiceCall.objects.create(
```
rudigiesler: One last thing, I'm wondering what we should do in case of duplicates? It might help to disallow duplicates for if an import fails halfway, and we need to retry it. Maybe we should change this to `get_or_create`?

vas2nets/models.py
erikh360: I avoided this incase they make changes.

rudigiesler: :+1:

erikh360:
@rudigiesler ready for another look.

vas2nets/tasks.py
```diff
@@ -0,0 +1,37 @@
...
+
+class FetchVoiceData(Task):
+
+    def get_data(self, date):
+        url = "%s?report_date=%s" % (settings.V2N_VOICE_URL, date)
+
+        content = requests.get(url, stream=True)
+
+        return csv.DictReader(
+            [line.decode('utf-8') for line in content.iter_lines()])
```
rudigiesler: `iter_lines` has a `decode_unicode` parameter for just this purpose. Maybe we can use that for cleaner code?

rudigiesler (commented):
Just one last small thing, otherwise looking great :)

Then I just thought of something. What if the API endpoint takes a date in its data, which defaults to yesterday? Then instead of writing a management command, you can just do a loop through the dates we want to backfill? It's just a thought, I'm happy with either one.

erikh360:
@rudigiesler thanks, not sure how I missed `decode_unicode`. I have already started with the history but I might adapt it with your suggestion, maybe even accept a date range, but in a separate PR.

rudigiesler (approved):

PR #170, opened 2017-09-28
API to populate call history

erikh360:


vas2nets/tasks.py
```diff
@@ -34,3 +35,26 @@ def run(self, date, **kwargs):
...
+
+        start_date = datetime.strptime(start_date, '%Y-%m-%d')
+        end_date = datetime.strptime(end_date, '%Y-%m-%d') + timedelta(1)
+
+        def daterange(start, end):
+            for n in range(int((end - start).days)):
+                yield start + timedelta(n)
+
+        for day in daterange(start_date, end_date):
+            if day.weekday() < 5:
```
rudigiesler: I'm not entirely sure why we're only fetching weekdays? Especially since Friday is a big send day, and it's possible that some sends might've gone past midnight in the past?

vas2nets/tasks.py
```diff
@@ -34,3 +35,26 @@ def run(self, date, **kwargs):
...
+                if VoiceCall.objects.filter(
+                        created_at__year=day.year,
+                        created_at__month=day.month,
+                        created_at__day=day.day).count() == 0:
```
rudigiesler: Since we're filtering, this makes retrying quite difficult, if a specific day failed processing half way.

Since we made sure that we're not creating duplicates, it is safe to just run this for every day.

rudigiesler (approved):

PR #172, opened 2017-10-10
[MOMNG-828] Add household phone number to registrations report

alexmuller:
This pull request is slightly speculative as I'm not completely familiar with the structure of the linked contacts/numbers yet. It probably needs careful review.


rudigiesler (approved):
:+1: Sorry for the slow review

PR #174, opened 2017-10-23
Fix missing gatekeeper for registration report generation

rudigiesler:
When generating the report, we use `','.join(msisdns)` to create the cell value for the list of msisdns. But when an identity doesn't have a gatekeeper, we default that value to None. This causes an error with the `join` function. This should default to `[]` so that the join function produces an empty string.


KaitCrawford (approved):
:+1:

PR #178, opened 2017-11-06
Fix operator identity lookup

erikh360:


registrations/tasks.py
```diff
@@ -468,7 +477,7 @@ def get_data(self):
     def create_registration(self, line, source):
         mother_identity = self.get_or_create_identity(
             line['mothers_phone_number'], details={})
-        operator_identity = self.get_or_create_identity(
+        operator_identity = self.get_operator_identity(
             line['health_worker_phone_number'])
```
rudigiesler: I'm very confused by this. Why is the `health_worker_phone_number` not a phone number, but a personnel code?

rudigiesler (commented):
Just one comment from me

registrations/tasks.py
erikh360: It's because I'm reusing the functionality that pulls registrations from a 3rd party which is not being used in Prod yet.

This should probably change to personnel code, but not really sure how to approach this.

rudigiesler (approved):

PR #179, opened 2017-11-07
Command to fix operator data from v2n

erikh360:
This is to link all the registrations created by the 3rd party api to the correct operators.

They were all created by v2n so I'm using source to identify them.


rudigiesler (approved):
I'm guessing that the validation on these registrations don't need to be rerun? If so, :+1:

erikh360:
No, we don't need to run the validation again, the validation checks that the `operator_id` field is there and a valid uuid and with this code that will always be valid.

PR #180, opened 2017-11-08
Change old personnel code lookup

erikh360:
I got the old personnel code from the incorrectly created identity using the `default_address` field. I made a incorrect assumption that this will always be populated.

The `msisdn` will always be populated so I'm using that instead.


KaitCrawford (approved):
:+1:

PR #182, opened 2017-11-10
[MOMNG-843] Start of pathfinder report

alexmuller:
This is the scaffold and start of the functionality to add a new kind of report for a Pathfinder cohort study.

Commit detail:

##  Remove registration-fixes.txt

I think this file is unused

## Cache container image after installing dependencies

This makes the build faster in development because we don't have to keep running `pip install`.

##  Add endpoint to list kinds of reports 

Could be used in the control-interface.

## Add start of new task for generating a message report 

Needs more work to fetch the details from the message-sender.

##  Add view method for creating new report 

`POST` to create a task which generates this new report.


reports/views.py
```diff
@@ -56,6 +61,40 @@ def post(self, request, *args, **kwargs):
...
+
+    POST: Start generating a report.
+    """
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request, *args, **kwargs):
+        request_data = request.data.copy()
+
+        request_data['start_date'] = '2017-09-01'
+        request_data['end_date'] = '2018-09-01'
```
KaitCrawford, line 76: I was thinking we could make this slightly more generic and the user could specify the dates. Would that complicate things?

KaitCrawford (commented):
looking good! Just one comment

KaitCrawford:
@alexmuller I think I'm going to land this PR and build on top of it. If I just continue with this on the same branch then this PR is going to get very bloated and I'd rather have the scaffold you've added in place

alexmuller:
@KaitCrawford I think that's probably easier, thanks!

PR #183, opened 2017-11-22
Task to generate message report for given msisdns

KaitCrawford:
This adds the functionality for the Pathfinder Cohort study

### DIFF
--- a/reports/tasks/msisdn_message_report.py
+++ b/reports/tasks/msisdn_message_report.py
@@ -182,14 +182,12 @@ class GenerateMSISDNMessageReport(BaseTask):
             'Message type'
         ]
 
-        i = 1
-        while i <= list_length:
+        for i in range(1, list_length + 1):
             header.extend([
                 'Message %d: content' % i,
                 'Message %d: date sent' % i,
                 'Message %d: status' % i
             ])
-            i += 1
 
         sheet.set_header(header)
 
@@ -207,12 +205,11 @@ class GenerateMSISDNMessageReport(BaseTask):
                 'Pregnancy week': data[msisdn].get('preg_week', ''),
                 'Message type': data[msisdn].get('msg_type', '')
             }
-            j = 1
-            for message in data[msisdn].get('messages', []):
-                row['Message %d: content' % j] = message['content']
-                row['Message %d: date sent' % j] = message['date_sent']
-                row['Message %d: status' % j] = message['status']
-                j += 1
+
+            for i, message in enumerate(data[msisdn].get('messages', []), 1):
+                row['Message %d: content' % i] = message['content']
+                row['Message %d: date sent' % i] = message['date_sent']
+                row['Message %d: status' % i] = message['status']
 
             sheet.add_row(row)
 
--- a/reports/tasks/msisdn_message_report.py
+++ b/reports/tasks/msisdn_message_report.py
@@ -13,7 +13,7 @@ class GenerateMSISDNMessageReport(BaseTask):
     MSISDNs.
     """
 
-    def run(self, task_status_id, msisdns=[]):
+    def run(self, start_date, end_date, task_status_id, msisdns=[]):
         task_status = ReportTaskStatus.objects.get(id=task_status_id)
         task_status.status = ReportTaskStatus.RUNNING
         task_status.save()
--- a/reports/tasks/msisdn_message_report.py
+++ b/reports/tasks/msisdn_message_report.py
@@ -148,7 +148,7 @@ class GenerateMSISDNMessageReport(BaseTask):
                     "content": message['content'],
                     "date_sent": datetime.strptime(message['created_at'],
                                                    "%Y-%m-%dT%H:%M:%S.%fZ"),
-                    "status": 'Delivered' if message['delivered'] else 'Undelivered'
+                    "status": 'Delivered' if message['delivered'] else 'Undelivered'  # noqa
                 })
 
             data[msisdn]["messages"] = message_list
@@ -189,7 +189,7 @@ class GenerateMSISDNMessageReport(BaseTask):
             row = {
                 'Phone number': msisdn,
                 'Date registered': data[msisdn].get('reg_date', ''),
-                'Facility': data[msisdn].get('facility', ""),
+                'Facility': data[msisdn].get('facility', ''),
                 'Pregnancy week': data[msisdn].get('preg_week', ''),
                 'Message type': data[msisdn].get('msg_type', '')
             }
--- a/reports/tasks/msisdn_message_report.py
+++ b/reports/tasks/msisdn_message_report.py
@@ -23,21 +23,21 @@ class GenerateMSISDNMessageReport(BaseTask):
         task_status.status = ReportTaskStatus.RUNNING
         task_status.save()
 
-        self.is_client = IdentityStoreApiClient(
+        is_client = IdentityStoreApiClient(
             settings.IDENTITY_STORE_TOKEN,
             settings.IDENTITY_STORE_URL,
         )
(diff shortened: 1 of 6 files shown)


reports/views.py
```diff
@@ -88,6 +85,8 @@ def post(self, request, *args, **kwargs):
         })
 
         generate_msisdn_message_report.apply_async(kwargs={
+            "start_date": datetime.strftime(data['start_date'], '%Y-%m-%d'),
+            "end_date": datetime.strftime(data['end_date'], '%Y-%m-%d'),
             "task_status_id": task_status.id,
             "msisdns": data['msisdns'],
```
rudigiesler, line 91: Do we not also need to pass the email addresses and details to the task, so that it sends the emails?

reports/tasks/msisdn_message_report.py
```diff
@@ -13,63 +16,200 @@ class GenerateMSISDNMessageReport(BaseTask):
...
 
-            sheet.add_row({
-                'Phone number': msisdn,
-                'Date registered': identity['created_at'],
+        return data
+
+    def retrieve_registration_info(self, is_client, data):
+        logger = self.get_logger()
+
+        for msisdn in data:
```
rudigiesler: Since each value in the data is a dictionary, we can probably do a `for msisdn, datum in data.items()`, and then when updating the values below, instead of doing the `data[msisdn]['key'] = 'value'`, we can just do `datum['key'] = 'value'`

reports/tasks/msisdn_message_report.py
```diff
@@ -13,63 +16,200 @@ class GenerateMSISDNMessageReport(BaseTask):
...
+                    'No operator_id on registration for {0}'.format(msisdn))
+                data[msisdn]['facility'] = ""
+
+        return data
+
+    def retrieve_messages(self, ms_client, data, start_date, end_date):
+        logger = self.get_logger()
+
+        longest_list = 0
+        for msisdn in data:
```
rudigiesler: Same here with iterating over `data.items()`

reports/tasks/msisdn_message_report.py
```diff
@@ -13,63 +16,200 @@ class GenerateMSISDNMessageReport(BaseTask):
...
+                    'No results from message sender for {0}'.format(msisdn))
+
+                data[msisdn]["messages"] = message_list
+                continue
+
+            for message in results:
+                message_list.append({
+                    "content": message['content'],
+                    # Reformat the date
+                    "date_sent": message['created_at'].replace('T', ' ')[:-8],
```
rudigiesler: This seems a little fragile to me. Maybe it's better to first get the string into python's `datetime` format, and then to `strftime` it to get it into the format that we want.

reports/tasks/msisdn_message_report.py
```diff
@@ -13,63 +16,200 @@ class GenerateMSISDNMessageReport(BaseTask):
...
+            file_name = 'msisdn-report-%s-to-%s.xlsx' % (
+                start_date.strftime('%Y-%m-%d'),
+                end_date.strftime('%Y-%m-%d'))
+            SendEmail.apply_async(kwargs={
+                'subject': email_subject,
+                'file_name': file_name,
+                'file_location': output_file,
+                'sender': email_sender,
+                'recipients': email_recipients,
+                'task_status_id': task_status_id})
```
rudigiesler, line 67: Do we maybe need to update the task status to completed here at the end?

reports/tasks/msisdn_message_report.py
```diff
@@ -13,63 +16,200 @@ class GenerateMSISDNMessageReport(BaseTask):
...
+        header = [
+            'Phone number',
+            'Date registered',
+            'Facility',
+            'Pregnancy week',
+            'Message type'
+        ]
+
+        i = 1
+        while i <= list_length:
```
rudigiesler: `for i in range(1, list_length + 1)`?

reports/tasks/msisdn_message_report.py
```diff
@@ -13,63 +16,200 @@ class GenerateMSISDNMessageReport(BaseTask):
...
+            header.extend([
+                'Message %d: content' % i,
+                'Message %d: date sent' % i,
+                'Message %d: status' % i
+            ])
+            i += 1
+
+        sheet.set_header(header)
+
+        for msisdn in msisdns:
```
rudigiesler: For any msisdn that doesn't have a related identity, we've set the value in `data` to an empty dictionary. So we can probably iterate over `data.items()` here, and we'll still get the full list of msisdns.

reports/tasks/msisdn_message_report.py
```diff
@@ -13,63 +16,200 @@ class GenerateMSISDNMessageReport(BaseTask):
...
+
+            row = {
+                'Phone number': msisdn,
+                'Date registered': data[msisdn].get('reg_date', ''),
+                'Facility': data[msisdn].get('facility', ''),
+                'Pregnancy week': data[msisdn].get('preg_week', ''),
+                'Message type': data[msisdn].get('msg_type', '')
+            }
+            j = 1
+            for message in data[msisdn].get('messages', []):
```
rudigiesler: What about `for i, message in enumerate(data[msisdn].get('messages', []), start=1)`?

reports/tests/test_msisdn_message_report.py
```diff
@@ -1,60 +1,679 @@
...
+        """
+        Generating a report should mark the ReportTaskStatus objects as Running
+        while the generation is in progress.
+        """
+
+        orig = openpyxl.Workbook.save
+
+        def new_save(wb, file):
+            task_status = ReportTaskStatus.objects.last()
+            self.assertEqual(task_status.status, ReportTaskStatus.RUNNING)
```
rudigiesler: Since we're doing the assert in here, if `openpyxl.Workbook.save` is never called, this test will pass successfully. I think we either need to assert that the mock was run, or have a variable that changes inside this function so that we can assert that this has been run

reports/tasks/msisdn_message_report.py
KaitCrawford: :man_facepalming:

rudigiesler: I also just saw that there is a test to check that the status is completed. Is that test somehow faulty? Or is there a task status change somewhere that I'm missing?

reports/tasks/msisdn_message_report.py
KaitCrawford: My reasoning for iterating over the list here rather than the dict is to maintain the order of the msisdns in the spreadsheet since the dict doesn't preserve the ordering.

rudigiesler: I'm not sure if the ordering is important or not, but it's not a big thing, so happy to keep it this way to maintain the ordering.

reports/views.py
KaitCrawford: :man_facepalming:

reports/tasks/msisdn_message_report.py
KaitCrawford: @rudigiesler this change happens in the send_email task

rudigiesler: Ah okay, but if there are no email addresses specified, then we should still mark the task as complete?

KaitCrawford: :+1:

reports/tests/test_msisdn_message_report.py
KaitCrawford: :face_with_head_bandage: :+1:

rudigiesler (approved):

PR #189, opened 2018-01-12
Add check for already subscribed mothers

rudigiesler:
For external registrations, we want to reject registrations for mothers that are already subscribed to an existing prebirth messageset.

### DIFF
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -395,6 +395,13 @@ class RepopulateMetrics(Task):
 repopulate_metrics = RepopulateMetrics()
 
 
+class AlreadySubscribedError(Exception):
+    """
+    For when a mother is already subscribed, but a new registration is being
+    requested
+    """
+
+
 class PullThirdPartyRegistrations(Task):
 
     def get_or_create_identity(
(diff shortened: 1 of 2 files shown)


dubesoftware (approved):
I see nothing amiss in the code, short of running it.

registrations/tasks.py
```diff
@@ -474,12 +481,32 @@ def get_data(self):
...
+        """
+        Ensures that the mother doesn't have any existing prebirth
+        subscriptions
+        """
+        subscriptions = utils.get_subscriptions(identity_id)
+        for subscription in subscriptions:
+            if (
+                    not subscription['active'] or subscription['completed'] or
+                    subscription['process_status'] not in (0, 1)):
+                pass
```
erikh360: Looks like this needs to be a `continue`, at the moment the check would happen regardless of subscription status.

erikh360: I think it would be a good idea to add a test where this `if` statement is True.

erikh360 (commented):
Looks good, just the one comment.

rudigiesler:
@erikh360 Thanks for that review, good catch. I've fixed the issue, ready for review again.

erikh360 (approved):
Yay!
Pending tests.

PR #190, opened 2018-02-27
Add task to sync welcome audio files over sftp

erikh360:


vas2nets/views.py
```diff
@@ -47,7 +47,6 @@ class SyncWelcomeAudioView(APIView):
         vas2nets sftp folder
     """
     def post(self, request, *args, **kwargs):
-        print 'post'
```
OMosimege: This looks great :) looks like the print line was meant to be deleted

vas2nets/views.py
```diff
@@ -37,3 +38,23 @@ def post(self, request, *args, **kwargs):
...
+
+
+class SyncWelcomeAudioView(APIView):
+
+    """ SyncWelcomeAudio Interaction
+        POST - starts up the task that sync the welcome audio files with the
+        vas2nets sftp folder
+    """
+    def post(self, request, *args, **kwargs):
+        print 'post'
```
OMosimege: This looks good cake. 
But like the previous review, there will be a few comments, and questions from my side just for learning experience

OMosimege: Do you want this print to be here? or for the method to just return the response

OMosimege (commented):
On a side note, is there a reason travis is failing?

vas2nets/views.py
erikh360: I had it there while testing, but I removed it.

erikh360:
@OMosimege you can click on the little red cross and it will show you the travis log, this one probably failed because flake8 picked up the `print` which I later removed. As long as the last commit passes.

Thanks for looking, please give both a thumbs up if you don't see anything wrong. 

Remember that it still need to go through QA so if you miss something it's not a issue.

vas2nets/views.py
OMosimege: Awesome, this looks good then

OMosimege (approved):

PR #194, opened 2018-04-12
Send welcome messages for public registrations

erikh360:
On completion of a public sign up, an SMS should be sent to the number(s) associated with the subscription to inform them that they have been registered to receive the HelloMama public message set.

### DIFF
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -238,7 +238,25 @@ class ValidateRegistration(Task):
         }
 
         # Add mother welcome message
-        if 'voice_days' in registration.data and \
+        if registration.stage == 'public':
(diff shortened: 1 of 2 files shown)


rudigiesler (approved):
Just to get this logic straight.

So if it's a public registration, then we want to send to both the mother and gatekeeper (or either or if the registration is for mother only or gatekeeper only)

If it's not a public registration, then we want to send only to the mother, or if it's a gatekeeper only registration, only to the gatekeeper.

If that logic is correct, then I'm 👍 on this PR

erikh360:
@rudigiesler Yes that is correct, thanks!

PR #196, opened 2018-04-25
Corp notification for non-full set registrations

erikh360:
I messed up the branch, it should be MOMNG-889.

This is to send notification to the corps on the subscribers they registered on the public messageset, If they complete the messageset and is not moved to the full one then a notification is sent.

The Corp gets a list of msisdns, it is triggered by a url so can be weekly or daily.

### DIFF
--- a/hellomama_registration/utils.py
+++ b/hellomama_registration/utils.py
@@ -108,6 +108,18 @@ def get_identity_address(identity):
     return identity_store_client.get_identity_address(identity)
 
 
+def get_address_from_identity(identity):
+    last_address = None
+    for address, detail in identity['details'].get(
+            'addresses', {}).get('msisdn', {}).items():
+        if detail.get('optedout', False) is True:
+            pass
+        if detail.get('default', False) is True:
+            return address
+        last_address = address
+    return last_address
+
+
 def search_identities(search_key, search_value):
     """
     Returns the identities matching the given parameters
@@ -178,6 +190,11 @@ def get_messageset(messageset_id):
     return stage_based_messaging_client.get_messageset(messageset_id)
 
 
+def search_messagesets(params):
+    r = stage_based_messaging_client.get_messagesets(params=params)
+    return r["results"]
+
+
 def get_schedule(schedule_id):
     return stage_based_messaging_client.get_schedule(schedule_id)
 
@@ -186,6 +203,12 @@ def get_subscriptions(identity):
     """ Gets the active subscriptions for an identity
     """
     params = {'identity': identity, 'active': True}
+    return search_subscriptions(params)
+
+
+def search_subscriptions(params):
+    """ Gets the subscriptions based on the params
+    """
     r = stage_based_messaging_client.get_subscriptions(params=params)
     return r["results"]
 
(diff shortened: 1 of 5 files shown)


registrations/tasks.py
```diff
@@ -652,3 +653,50 @@ def run(self, user_id, **kwargs):
...
+class SendPublicRegistrationNotifications(Task):
+    """
+    Send out a notification SMS to the CORP for public registrations not
+    converted to full registrations
+    """
+
+    def run(self):
+
+        messagesets = utils.search_messagesets(
+            {'short_name__contains': 'public'})
```
rudigiesler: I don't think that there is a filter that will filter these messagesets by short name contains. We'll probably have to do the filtering client side.

registrations/tasks.py
```diff
@@ -652,3 +653,50 @@ def run(self, user_id, **kwargs):
...
+        messageset_ids = [str(messageset['id']) for messageset in messagesets]
+
+        subscriptions = utils.search_subscriptions(
+            {'completed': True, 'metadata_not_has_key': 'public_notification',
+             'messageset__in': ','.join(messageset_ids)})
```
rudigiesler: Same here with not being able to filter the subscriptions by `messageset__in` or `metadata_not_has_key`

hellomama_registration/utils.py
```diff
@@ -108,6 +108,18 @@ def get_identity_address(identity):
...
+def get_address_from_identity(identity):
+    last_address = None
+    for address, detail in identity['details'].get(
+            'addresses', {}).get('msisdn', {}).items():
+        if detail.get('optedout', False) is True:
+            pass
+        if detail.get('default', False) is True:
+            return address
+        last_address = address
+    return last_address
```
rudigiesler, line 120: Instead of using this, we could use the newer `/identities/{uuid}/addresses/{address_type}` endpoint, with the `default=true` parameter.

registrations/tasks.py
```diff
@@ -652,3 +653,50 @@ def run(self, user_id, **kwargs):
...
+
+            utils.patch_subscription(
+                subscription, {"public_notification": "true"})
+
+        for corp_id, msisdns in corp_details.items():
+            payload = {
+                "to_identity": corp_id,
+                "content":
+                    "Public registrations not on full set: {}".format(
+                        ', '.join(msisdns)),
```
rudigiesler: If this is a very large list of msisdns, then this SMS might be too many parts and would fail to send. We should probably put some sort of limit and batch it into multiple SMSes.

registrations/tasks.py
erikh360: I added the `metadata_not_has_key` here https://github.com/praekelt/seed-stage-based-messaging/pull/113 I'll just double check the others, thanks for the review.

hellomama_registration/utils.py
erikh360: I do it this way to avoid doing another call to the identity store, before calling this function I retrieve the whole identity to get the operator id to know where to send the notification.

rudigiesler (approved):

PR #203, opened 2018-05-18
Add key for reporting

erikh360:
We need to know if a public subscription was converted to the full set for the following reports:
% conversion from Public to Full Message set total / per month 
% conversation from Public to Full Message set per weekly reminder (1/2/3/4).

I am assuming that if a public subscription is complete and then a full subscription is created that it is not a conversion. <- This was confirmed.

### DIFF
--- a/registrations/tasks.py
+++ b/registrations/tasks.py
@@ -216,8 +216,14 @@ class ValidateRegistration(Task):
                 {"identity": registration.mother_id, "completed": False,
                  "active": True, "messageset_contains": "public.mother"})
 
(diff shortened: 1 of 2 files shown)


rudigiesler (approved):
Looks good :)

The way it is currently, if there's a mom + household subscription, when they register for full messages, it will count as multiple conversions. Is this how we want it to reflect in the stats?

erikh360:
@rudigiesler we would be able to filter on the message set and get mother conversions and household conversions separately.

PR #204, opened 2018-06-11
Personnel csv import

erikh360:
Currently we are importing personnel details using the cli, there is no validation using this method and the data is very inconsistent. This makes reporting very hard.

We can now maintain valid states, facilities and communities and validate against them when we import new personnel/CORPs.


uniqueids/admin.py
```diff
@@ -27,4 +31,128 @@ def resend_personnel_code(self, request, queryset):
...
+
+        required_keys_type = {
+            "corps": ["community"],
+            "personnel": ["role", "facility_name", "state"]
+        }
+
+        missing = []
+        for key in required_keys + required_keys_type[import_type]:
+            if key not in record.keys():
+                missing.append(key)
```
rudigiesler: If we use `set`s here, then it can be done a bit easier with set arithmetic, ie
```
>>> required = set(['a', 'b', 'c'])
>>> record = set(['a', 'b'])
>>> required - record
set(['c'])
```

uniqueids/admin.py
```diff
@@ -27,4 +31,128 @@ def resend_personnel_code(self, request, queryset):
...
+            obj.valid = False
+        else:
+            for line in rows:
+                missing_keys = self.validate_keys(line, obj.import_type)
+
+                if missing_keys:
+                    for key in missing_keys:
+                        if key not in missing_fields:
+                            missing_fields.append(key)
+                    obj.valid = False
```
rudigiesler: If we use sets here too, then this logic becomes a lot easier too (along with missing, states, facilities and communities):
```
>>> s = set()
>>> s.add('a')
>>> s.add('a')
>>> s
set(['a'])
```

uniqueids/models.py
```diff
@@ -30,6 +30,61 @@ def __str__(self):
...
+    def __str__(self):
+        return self.name
+
+
+@python_2_unicode_compatible
+class PersonnelUpload(models.Model):
+    PERSONNEL_CHOICES = (
+        ("personnel", "Personnel"),
+        ("corps", "CORPs")
+    )
```
rudigiesler: Usually we would do this more like:
```
PERSONNEL_TYPE = 'p'
CORP_TYPE = 'c'
PERSONNEL_CHOICES = (
    (PERSONNEL_TYPE, "Personnel"),
    (CORP_TYPE, "CORPs")
)
```

That way, it takes up less space in the DB, as well as when you're referring to it in the code in other places, you can do `PersonnelUpload.CORP_TYPE` and get autocomplete and earlier error reporting, so that it's much less likely to make a type than using the string directly.

erikh360:
@rudigiesler This is ready for another look.

rudigiesler (approved):

PR #206, opened 2018-06-12
Fix payload for identity create

erikh360:


DevChima:
+1

uniqueids/admin.py
```diff
@@ -140,7 +140,23 @@ def save_model(self, request, obj, form, change):
 
         if obj.valid:
             for line in rows:
-                utils.create_identity(line)
+                identity = {
+                    "communicate_through": line.get("communicate_through"),
+                    "details": {
+                        "addresses": {
+                            line["address_type"]: {
+                                line["address"]: {}
```
rudigiesler, line 148: Not sure if you do it for hellomama as well, but for momconnect we do a `"default": true` in the details of the address, ie
```
"addresses": {
  "+27821234567": {"default": true}
}
```

erikh360: Looking at some of the latest identities I see that we need that, I'll create another PR.

PR #209, opened 2018-06-13
Add values and existing address check

erikh360:


uniqueids/admin.py
```diff
@@ -54,22 +54,59 @@ class PersonnelUploadAdmin(admin.ModelAdmin):
...
+            elif key == "address_type" and value != "msisdn":
+                missing.add(key)
+            elif (key == "preferred_language" and
+                    value not in settings.LANGUAGES):
+                missing.add(key)
+            elif (key == "uniqueid_field_name" and
+                    value not in ['personnel_code', 'corp_code']):
+                missing.add(key)
+            elif (key == "uniqueid_field_length" and not value.isdigit() and
+                    value.find('-') == -1):
```
rudigiesler: I'm not sure why we need the `value.find('-')`, surely `isdigit` will filter out any values with `-`'s in them?

uniqueids/admin.py
```diff
@@ -54,22 +54,59 @@ class PersonnelUploadAdmin(admin.ModelAdmin):
...
-        required = set(required_keys + required_keys_type[import_type])
+    def validate_keys(self, record, import_type):
+        required = set(
+            self.required_keys + self.required_keys_type[import_type])
         missing = required - set(record.keys())
 
         return missing
 
+    def validate_values(self, record, import_type):
+        missing = set()
```
rudigiesler: I don't quite think that `missing` is the correct name for this variable. Maybe something like `invalid` is better, especially since we're assigning this result to an `invalid_values` variable later on.

uniqueids/admin.py
```diff
@@ -54,22 +54,59 @@ class PersonnelUploadAdmin(admin.ModelAdmin):
...
+        identities = utils.search_identities(
+            "details__addresses__{}".format(record["address_type"]), msisdn)
+
+        for key in identities:
+            return True
+
+        if len(msisdn) != 13:
+            return True
+
+        return False
```
rudigiesler: At first look this function is a bit confusing, because it's returning `True` if the address is invalid, and `False` if the address is valid. Maybe we can switch those two around, and `not` the result when we use it below?

uniqueids/admin.py
```diff
@@ -100,45 +139,61 @@ def save_model(self, request, obj, form, change):
...
+                if ("address" in line and "address_type" in line):
+                    if (self.validate_address(line)):
+                        existing_address.add(line["address"])
+
                 missing_keys = self.validate_keys(line, obj.import_type)
 
                 if missing_keys:
                     for key in missing_keys:
                         missing_fields.add(key)
-                    obj.valid = False
```
rudigiesler: This block can probably be simplified by `missing_fields |= self.validate_keys(line, obj.import_type)`

uniqueids/admin.py
```diff
@@ -100,45 +139,61 @@ def save_model(self, request, obj, form, change):
...
+
+                missing_keys = self.validate_values(line, obj.import_type)
+
+                if missing_keys:
+                    for key in missing_keys:
+                        invalid_values.add(key)
```
rudigiesler: We seem to be overwriting `missing_keys` here, which is probably bad, and also a bit misleading as to what that variable now contains.

We can also simplify this to `invalid_values |= self.validate_values(line, obj.import_type)`

uniqueids/admin.py
```diff
@@ -100,45 +139,61 @@ def save_model(self, request, obj, form, change):
...
+                errors.append(
+                    "Address invalid or already exists: {}".format(
+                        ', '.join(sorted(existing_address))))
             if missing_fields:
                 errors.append("Missing fields: {}".format(', '.join(
-                    missing_fields)))
+                    sorted(missing_fields))))
+            if missing_keys:
+                errors.append("Missing or invalid values: {}".format(', '.join(
+                    sorted(missing_keys))))
```
rudigiesler: This should probably be for `missing_fields` and `missing_values`, not `missing_keys` which is a variable that is constantly overwritten in the loop

uniqueids/admin.py
erikh360: I'm just checking for a positive number here, `isdigit` still returns True for a negative.

rudigiesler: It doesn't seem to be according to the docs: https://docs.python.org/3/library/stdtypes.html#str.isdigit , or my brief testing:
```
>>> '1'.isdigit()
True
>>> '-1'.isdigit()
False
```

erikh360: I was convinced I tried that locally with different results. Will update now.

erikh360:
@rudigiesler ready for another look.

rudigiesler (approved):

PR #211, opened 2018-06-18
Fix address validation

erikh360:


uniqueids/tests.py
```diff
@@ -760,7 +760,8 @@ def test_personnel_upload_good(self):
             {
                 'communicate_through': None,
                 'details': {
-                    'addresses': {'msisdn': {'+2347012312341': {"default": True}}},
+                    'addresses': {
```
OMosimege: I'm assuming these were the missing environment variables that caused the previous fail?

OMosimege (approved):
Looks great Erik! 💯

OMosimege (approved):
everything looks great 💯

uniqueids/tests.py
erikh360: It was `flake8` complaining that the line was too long.

OMosimege: okay great 👍

PR #214, opened 2018-07-20
Endpoint for user details

erikh360:


registrations/migrations/0009_create_get_registrations_view.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('registrations', '0008_auto_20180323_1228'),
+    ]
+
+    operations = [
+        migrations.RunSQL(sql)
```
rudigiesler: I see that we're not providing a reverse function here. Maybe a simple `DROP FUNCTION IF EXISTS` would be good to put in, so that reversing of migrations isn't broken?

registrations/views.py
```diff
@@ -273,3 +275,54 @@ def post(self, request, *args, **kwargs):
...
+            columns = [col[0] for col in cursor.description]
+            return [
+                dict(zip(columns, row))
+                for row in cursor.fetchall()
+            ]
+
+        sql = """
+            select *
+            from get_registrations('{}', '*', '*', '*', '*', {}, {})""".format(
+            settings.DBLINK_CONN, page_size, offset)
```
rudigiesler: So we do set `DBLINK_CONN` ourselves, and I know that the user input `page` is forced into an int, but I'm still concerned for maybe in future if someone else comes along and changes things, that there's a chance for sql injection.

Instead of `.format` here, could we maybe not rather use the `params` argument of `execute`, which does protect against sql injection?

hellomama_registration/settings.py
```diff
@@ -356,5 +356,7 @@
 V2N_FTP_PASS = os.environ.get('V2N_FTP_PASS')
 V2N_FTP_ROOT = os.environ.get('V2N_FTP_ROOT')
 
+DBLINK_CONN = ''
```
rudigiesler: Should this not get the value from `os.environ` so that we can specify it in our config?

erikh360:
@rudigiesler thanks for picking up these. Ready for another look.

rudigiesler (approved):

PR #218, opened 2018-08-30
Missed call trigger sms

DevChima:
Removed special character in contraction and replaced with 'we will'


erikh360 (approved):
Cool! Happy when the tests pass.

